PR #39, opened 2021-01-03
fix: monitor['h'] statistical errors

chiiyuu:
The original version missed the iterative process of monitor['h'].


fangwei123456:
Thanks!

PR #201, opened 2022-04-02
Update adaptive neuron.

ridgerchu:


fangwei123456:
Store new codes in branch `neuron` before cupy backend with v_rest is finished.

Issue #361, opened 2023-03-31
Issue in SHD dataset when trying to use integrate by fixed duration


Thvnvtos:
The issue is in spikingjelly.datasets.shd
specifically in : integrate_events_by_fixed_duration_shd(events: Dict, duration: int, W: int)

    x = events['x']
    t = events['t']
    N = t.size

    frames = []
    left = 0
    right = 0
    while True:
        t_l = t[left]
        while True:
            if right == N or t[right] - t_l > duration:
                break
            else:
                right += 1

The values of events['t'] are in seconds (s)
`array([0.      , 0.001833, 0.002167, ..., 0.698   , 0.6997  , 0.7007  ],
      dtype=float16)`

Since the duration parameter is specified to be an int, we can't use this function correctly.
Using a duration that is in milliseconds (ms) yields wrong data like this:

![output](https://user-images.githubusercontent.com/23236210/229148268-d6b2cbdd-7ac3-4d70-bd85-1ed20365a74f.png)

A simple solution is to convert events['t'] to milliseconds (ms):

` t = 1000 * events['t']`

which results in correct integration by duration:

![output](https://user-images.githubusercontent.com/23236210/229152394-546f586b-4afe-4da8-92d7-06b33bd6c850.png)

I've opened a pull request for this change, if it looks ok.

fangwei123456:
Thanks! I have merged your PR.

fangwei123456:
This is a fatal bug, and I had added it to the bug report in https://github.com/fangwei123456/spikingjelly/blob/master/bugs.md.

PR #362, opened 2023-03-31
convert events['t'] from s to ms and add SpikingSpeechCommand

Thvnvtos:
1st commit: related to this issue: https://github.com/fangwei123456/spikingjelly/issues/361#issue-1649496882

2nd commit:
SpikingSpeechCommand (SSC) is part of the shd(Spiking Heidelberg Datasets) along with SpikingHeidelbergDigits(SHD) .
https://zenkelab.org/resources/spiking-heidelberg-datasets-shd/

I added SSC which is very similar to SHD, the only paramter that is different is train: bool, changed to split: str, for 'train', 'valid' and 'test'. Also SSC has more classes.

I tested this code and seems to be working fine.


fangwei123456:
Thanks!

PR #584, opened 2024-10-08
Replace np.int with int for compatibility with NumPy 1.20+

Doudou-Wu:
### Description
Replaced  `np.int` with Python's built-in `int` to address the deprecation warning in NumPy 1.20+.

### Issue
Fixes compatibility issue with NumPy 1.20+, which deprecated `np.int`.

### Testing
All tests were run successfully, and no existing functionality was broken.

### DIFF
--- a/spikingjelly/activation_based/auto_cuda/base.py
+++ b/spikingjelly/activation_based/auto_cuda/base.py
@@ -246,7 +246,7 @@ class CKernel:
                 elif value.dtype == np.float16:
                     assert startswiths(ctype, ('const half2', 'half2'))
 
-                elif value.dtype == np.int_:
+                elif value.dtype == int_:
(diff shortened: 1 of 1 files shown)


fangwei123456:
This change will cause an error:
```
Python 3.9.20 (main, Oct  3 2024, 07:27:41) 
[GCC 11.2.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> int_
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'int_' is not defined
```

fangwei123456:
I fixed it in: https://github.com/fangwei123456/spikingjelly/commit/5e95cddc25962e5119682711defec3f020910bba

PR #618, opened 2025-05-01
improved path tracing for naming

chanokin:
I was trying to export and save converted SNN to later load the `state_dict` but the naming of the injected `VoltageScalers` and `LINodes` was a bit hard to work with. I think I managed to get a good compromise for the replacement naming, what do you think?


fangwei123456:
@Lyu6PosHao

PR #676, opened 2026-05-14
Refactor op_counter.neuromc

AllenYolk:
This PR tightens SpikingJelly's NeuroMC runtime profiler and aligns it more closely with the current exact-model implementation.

The main changes are:
- promote memory residency tracking into a generic `op_counter.memory_residency` counter
- keep `op_counter.neuromc` focused on exact NeuroMC-style runtime energy estimation
- fix several correctness issues found during review
- refresh the NeuroMC runtime tests around the exact profiler behavior

## Summary by CodeRabbit

* **New Features**
  * Added memory‑residency profiling and a new exact runtime energy profiler with optimizer‑step accounting, richer energy/mapping breakdowns, and model/optimizer binding.

* **Documentation**
  * Standardized and corrected docstrings, examples, and optional parameter/type annotations.

* **Refactor**
  * Consolidated residency implementations, simplified package exports, and redesigned memory‑hierarchy presets.

* **Bug Fixes**
  * Improved empty‑tensor handling and corrected docstring parameter metadata.

* **Tests**
  * Expanded test suite to validate exactness, stage/optimizer accounting, residency behavior, and error conditions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/fangwei123456/spikingjelly/pull/676)


spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -142,723 +210,1208 @@ def _resolve_loss_fn(loss_fn: Callable | None):
...
+    def bind_optimizer(self, optimizer: torch.optim.Optimizer | None):
+        self._optimizer = optimizer
 
     def __enter__(self):
         self._trace_mode.__enter__()
-        self._dispatch_mode.__enter__()
         self._active = True
         return self
 
     def __exit__(self, exc_type, exc, tb):
```
AllenYolk, line 426: ⚠️ **Warning:** `_neuromc_last_input` monkey-patches Conv/Linear modules but `__exit__` never cleans it up. After profiling, every bound module retains a dangling `_neuromc_last_input` attribute holding a stale tensor reference.

Suggestion: add cleanup in `__exit__`:
```python
for module in (self._bound_model.modules() if self._bound_model else []):
    if hasattr(module, '_neuromc_last_input'):
        del module._neuromc_last_input
```

This prevents memory leaks and attribute pollution on shared modules.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -53,59 +56,72 @@
     "aten.alias",
     "aten._unsafe_view",
     "aten.as_strided",
+    "aten.clone",
+    "aten.copy_",
+    "profiler.",
 )
 
+_AUXILIARY_ATEN_OPS = {
```
AllenYolk, line 66: ⚠️ **Warning:** `_AUXILIARY_ATEN_OPS` now includes `aten.add.Tensor`, `aten.add_.Tensor`, `aten.mul.Tensor`, `aten.mul_.Tensor` etc. — ops that the **old** profiler explicitly counted as primitives with per-op energy costs (0.548 pJ/add, 0.812 pJ/mul).

These are now silently treated as free. Is this intentional? If add/mul inside fused kernels are truly zero-cost in the exact NeuroMC model, this is fine — but it should be documented. If not, these ops should be removed from `_AUXILIARY_ATEN_OPS` so the trace fallback can still count them.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -53,59 +56,72 @@
...
+    "aten.logical_or.default",
+    "aten.logical_xor.default",
+    "aten.logical_not.default",
+    "aten.new_empty_strided.default",
+    "aten.ones_like.default",
+    "aten.zeros_like.default",
+}
 
 @dataclass
 class NeuroMCRuntimeEnergyReport:
```
AllenYolk, line 131: 🔴 **Critical — Docstring regression:** `NeuroMCRuntimeEnergyReport` is a public API dataclass that previously had comprehensive bilingual (CN/EN) docstrings explaining every field. Now it has zero documentation — just bare field declarations.

The same is true for `NeuroMCEnergyProfiler`, `get_report()`, `estimate_neuromc_runtime_energy()`, and `MemoryHierarchyConfig`. The project's AGENTS.md requires bilingual docstrings for all public APIs.

At minimum, add back field-level `:param:` docstrings or a class-level bilingual description explaining the new report structure (energy_mac_pj, energy_by_core_type, energy_by_process_key, etc.).

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -142,723 +210,1208 @@ def _resolve_loss_fn(loss_fn: Callable | None):
...
+            output_numel=kt + fyfxkc,
+            mac_count=0,
+            source="optimizer",
         )
 
-        energy_memory_pj = 0.0
-        for level, bits in level_bits.items():
-            energy_memory_pj += bits * self.memory_config.energy_pj_per_bit.get(
-                level, 0.0
+    def get_report(self) -> NeuroMCRuntimeEnergyReport:
```
AllenYolk, line 1573: ⚠️ **Warning — Missing docstring:** `get_report()` is the primary output method of the profiler. It previously had a comprehensive bilingual docstring. Now it has none.

Consider restoring a docstring that documents:
- The fragment-based analysis approach (hooks vs trace fallback)
- When `ValueError` is raised (unsupported ops, no fragments)
- The relationship between the new fields (energy_mac_pj, energy_base_memory_pj, energy_extra_memory_pj, etc.)

spikingjelly/activation_based/op_counter/memory_residency.py
```diff
@@ -0,0 +1,515 @@
...
+    tensors = _collect_tensors((args, kwargs, out))
+    has_grad_tensor = any(t.requires_grad for t in tensors)
+    is_optimizer_like = any(hint in op_name for hint in _OPTIMIZER_HINTS)
+    if has_grad_tensor and is_optimizer_like:
+        return "optimizer"
+
+    return "forward"
+
+
+def _extract_tensors(tree: Any) -> list[torch.Tensor]:
```
AllenYolk: 💡 **Suggestion:** `_collect_tensors` (line 46) and `_extract_tensors` (line 68) are identical functions — both flatten a pytree and extract tensors. Consolidate to one name and remove the other.

AllenYolk:
## Code Review Summary

**Verdict: Request Changes** (2 issues need attention, 3 suggestions)

---

### 🔴 Critical

- **core.py:124** — All bilingual docstrings (CN/EN) stripped from public APIs: `NeuroMCRuntimeEnergyReport`, `NeuroMCEnergyProfiler`, `estimate_neuromc_runtime_energy`, `MemoryHierarchyConfig`. This violates the project's AGENTS.md documentation requirements. The new report structure has 10+ new fields (`energy_mac_pj`, `energy_by_core_type`, `energy_by_process_key`, `mapping_summary`, etc.) that are completely undocumented. **Must restore before merge.**

### ⚠️ Warnings

- **core.py:312** — `__exit__` does not clean up `_neuromc_last_input` attributes monkey-patched onto Conv/Linear modules during `bind_model()`. After profiling, modules retain stale tensor references — memory leak and attribute pollution on shared modules.
- **core.py:64** — `_AUXILIARY_ATEN_OPS` includes `aten.add.Tensor` / `aten.mul.Tensor` etc., which the old profiler explicitly counted as primitives with energy costs (0.548 pJ/add, 0.812 pJ/mul). These are now silently free. If intentional in the exact model, document it; if not, remove from the set.
- **Breaking API change** — Removed parameters from public functions: `op_cost_pj`, `memory_cost_pj_per_bit`, `memory_level_weights`, `memory_model` from `estimate_neuromc_runtime_energy()`; `memory_model` field from `MemoryHierarchyConfig`. Existing downstream code passing these will get `TypeError`. This needs a deprecation path or migration guide.

### 💡 Suggestions

- **memory_residency.py:68** — `_collect_tensors` (line 46) and `_extract_tensors` (line 68) are identical. Consolidate.
- **core.py** — `_is_spike_tensor` (core.py:125) duplicates `_is_spike` (utils.py:69). Consider importing from utils or consolidating.
- **core.py:1255** — `get_report()` lost its docstring. It now has complex behavior (hook vs trace fallback, `ValueError` on unsupported ops) that deserves documentation.
- **core.py** — The `_fallback_fragments_from_trace` method is ~120 lines of deeply nested if/elif. Consider extracting helper methods for the gemm backward and conv forward paths.

### ✅ Looks Good

- Clean architectural separation: memory residency counter promoted to top-level, NeuroMC focused on exact energy modeling.
- The fragment-based approach (module hooks → `_Fragment` dataclass → energy computation) is a significant improvement over the old dispatch-counter pattern.
- `base_counter.py` graceful handling of unknown ops (return 0) is a good fix.
- `add_counter.py` fix for `_bias_sizes` and the `isinstance(out, (tuple, list))` guard are correct.
- Test coverage is solid for the new exact model — spike detection, batch norm, training flow with bp_grad/wg/optimizer, weight reuse, frozen params, error paths.
- The `MemoryInstanceSpec` + `MemoryHierarchyConfig` redesign with hardware-specific instances is well-structured.

---
*Reviewed by Hermes Agent • [PR #676](https://github.com/fangwei123456/spikingjelly/pull/676)*

AllenYolk (commented):
## Code Review Summary

**Verdict: Request Changes** (2 issues need attention, 3 suggestions)

---

### 🔴 Critical

- **core.py:124** — All bilingual docstrings (CN/EN) stripped from public APIs: `NeuroMCRuntimeEnergyReport`, `NeuroMCEnergyProfiler`, `estimate_neuromc_runtime_energy`, `MemoryHierarchyConfig`. This violates the project's AGENTS.md documentation requirements. The new report structure has 10+ new fields (`energy_mac_pj`, `energy_by_core_type`, `energy_by_process_key`, `mapping_summary`, etc.) that are completely undocumented. **Must restore before merge.**

### ⚠️ Warnings

- **core.py:312** — `__exit__` does not clean up `_neuromc_last_input` attributes monkey-patched onto Conv/Linear modules during `bind_model()`. After profiling, modules retain stale tensor references — memory leak and attribute pollution on shared modules.
- **core.py:64** — `_AUXILIARY_ATEN_OPS` includes `aten.add.Tensor` / `aten.mul.Tensor` etc., which the old profiler explicitly counted as primitives with energy costs (0.548 pJ/add, 0.812 pJ/mul). These are now silently free. If intentional in the exact model, document it; if not, remove from the set.
- **Breaking API change** — Removed parameters from public functions: `op_cost_pj`, `memory_cost_pj_per_bit`, `memory_level_weights`, `memory_model` from `estimate_neuromc_runtime_energy()`; `memory_model` field from `MemoryHierarchyConfig`. Existing downstream code passing these will get `TypeError`. This needs a deprecation path or migration guide.

### 💡 Suggestions

- **memory_residency.py:68** — `_collect_tensors` (line 46) and `_extract_tensors` (line 68) are identical. Consolidate.
- **core.py** — `_is_spike_tensor` (core.py:125) duplicates `_is_spike` (utils.py:69). Consider importing from utils or consolidating.
- **core.py:1255** — `get_report()` lost its docstring. It now has complex behavior (hook vs trace fallback, `ValueError` on unsupported ops) that deserves documentation.
- **core.py** — The `_fallback_fragments_from_trace` method is ~120 lines of deeply nested if/elif. Consider extracting helper methods for the gemm backward and conv forward paths.

### ✅ Looks Good

- Clean architectural separation: memory residency counter promoted to top-level, NeuroMC focused on exact energy modeling.
- The fragment-based approach (module hooks → `_Fragment` dataclass → energy computation) is a significant improvement over the old dispatch-counter pattern.
- `base_counter.py` graceful handling of unknown ops (return 0) is a good fix.
- `add_counter.py` fix for `_bias_sizes` and the `isinstance(out, (tuple, list))` guard are correct.
- Test coverage is solid for the new exact model — spike detection, batch norm, training flow with bp_grad/wg/optimizer, weight reuse, frozen params, error paths.
- The `MemoryInstanceSpec` + `MemoryHierarchyConfig` redesign with hardware-specific instances is well-structured.

---
*Reviewed by Hermes Agent • [PR #676](https://github.com/fangwei123456/spikingjelly/pull/676)*

AllenYolk:
@coderabbitai please review the latest validated update on this PR. @gemini-code-assist please review the latest validated update as well.

spikingjelly/activation_based/op_counter/neuromc/config.py
```diff
@@ -2,222 +2,152 @@
...
+            "sram_bp_grad_in_smask",
+            "sram_bp_grad_out_du",
+            "sram_wg_conv_in_s",
+            "sram_wg_conv_in_du",
+            "sram_wg_conv_out_dw",
+            "sram_2MB",
+            "sram_6MB",
+            "dram",
+        }
+        missing = sorted(required.difference(self.memory_instances.keys()))
```
AllenYolk, line 160: 🔴 **Bug:** `sram_bp_conv_in_w` is defined in the preset (L48) and used in `_base_memory_for_fragment` in `core.py`, but it's **missing from the `required` set**. A custom `memory_instances` dict that omits this key will pass validation but crash at runtime with a `KeyError`.

Fix: add `"sram_bp_conv_in_w"` to the `required` set.

spikingjelly/activation_based/op_counter/memory_residency.py
```diff
@@ -0,0 +1,521 @@
...
+            if bits > old:
+                self.usage_bits["sram"] += bits - old
+                self.sram_cache[key] = bits
+            self._touch(self.sram_cache, key)
+            return True
+
+        while self.sram_cache and (
+            self.usage_bits["sram"] + bits > self.capacity_bits["sram"]
+        ):
+            evict_key, evict_bits = self.sram_cache.popitem(last=False)
```
AllenYolk, line 296: 🔴 **Bug — silent data loss in move accounting.** When SRAM eviction removes a key that also exists in REG (`reg_cache.pop(evict_key, None)` on L282), no `_record_move` is called for the REG->DRAM traffic. Compare with `_insert_reg` which correctly records moves on eviction.

This means `move_bits_by_edge["reg->dram"]` and `move_bits_by_op[...]["reg->dram"]` are **undercounted** whenever data resides in both REG and SRAM when the SRAM entry is evicted.

Fix: after popping from `reg_cache`, add: `if reg_bits is not None: self.usage_bits["reg"] -= reg_bits; self._record_move("reg", "dram", reg_bits, op_name)`

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -142,725 +210,1229 @@ def _resolve_loss_fn(loss_fn: Callable | None):
...
-        :type strict: bool
-
-        :param verbose: verbose flag forwarded to ``DispatchCounterMode``
-        :type verbose: bool
-
-        :param extra_ignore_modules: extra modules to ignore
-        :type extra_ignore_modules: list[nn.Module]
-        """
+        if core_type not in _ALLOWED_CORE_TYPES:
+            raise ValueError(
```
AllenYolk, line 357: ⚠️ **Breaking API change.** The old code logged a warning for unsupported ops; this raises `ValueError`. Users with models containing `torch.sin`, `torch.tanh`, etc. will get hard crashes instead of degraded reports.

Consider adding a `strict: bool = True` parameter -- when `False`, log a warning and continue. This preserves backward compatibility while allowing strict mode for CI.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -27,18 +21,27 @@
...
+    "comp": 0.056,
     "sqrt": 0.514,
-    "mux": 0.548 * (1.0 / 16.0),
 }
 
-_DEFAULT_MEMORY_LEVEL_WEIGHTS = {
-    "reg": 2.0,
-    "sram": 1.0,
-    "dram": 0.25,
+_MAC_COST_PJ = {
```
AllenYolk, line 43: ⚠️ `_MAC_COST_PJ` defines costs for 3 core types (`fp_soma`, `bp_grad`, `wg`) but `_ALLOWED_CORE_TYPES` has 6. The `get()` default of `0.0` silently zeroes MAC energy for `fp_bn`, `bp_bn`, and `bp_grad_opt`.

If these core types genuinely produce no MACs, add an explicit comment documenting that. If they can produce MACs, this is an energy undercount.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -114,24 +130,76 @@ class NeuroMCRuntimeEnergyReport:
...
+
+
+def _prod(values) -> int:
+    p = 1
+    for v in values:
+        p *= int(v)
+    return p
+
+
+def _is_spike_tensor(x: Any) -> bool:
```
AllenYolk: 💡 `_is_spike_tensor` is nearly identical to `_is_spike` in `utils.py` -- same logic, only difference is `logical_or` (non-inplace) vs `logical_or_` (inplace). The inplace variant mutates the input tensor, which is risky.

Suggestion: fix `_is_spike` in utils.py to use non-inplace `logical_or`, then import it here instead of duplicating.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -142,725 +210,1229 @@ def _resolve_loss_fn(loss_fn: Callable | None):
...
+    def bind_optimizer(self, optimizer: torch.optim.Optimizer | None):
+        self._optimizer = optimizer
 
     def __enter__(self):
         self._trace_mode.__enter__()
-        self._dispatch_mode.__enter__()
         self._active = True
         return self
 
     def __exit__(self, exc_type, exc, tb):
```
AllenYolk, line 426: ⚠️ If `handle.remove()` or the `delattr` loop raises, `self._trace_mode.__exit__` is never called, leaving trace mode active. Wrap cleanup in try/finally so the trace mode teardown always runs.

spikingjelly/activation_based/op_counter/memory_residency.py
```diff
@@ -0,0 +1,521 @@
...
+    tensors = _collect_tensors((args, kwargs, out))
+    has_grad_tensor = any(t.requires_grad for t in tensors)
+    is_optimizer_like = any(hint in op_name for hint in _OPTIMIZER_HINTS)
+    if has_grad_tensor and is_optimizer_like:
+        return "optimizer"
+
+    return "forward"
+
+
+def _extract_tensors(tree: Any) -> list[torch.Tensor]:
```
AllenYolk: 💡 `_collect_tensors` (L46) and `_extract_tensors` (L68) are **identical implementations**. Consolidate into one function and update the call sites.

spikingjelly/activation_based/op_counter/memory_residency.py
```diff
@@ -0,0 +1,521 @@
...
+from torch.utils._pytree import tree_flatten
+
+from .base import BaseCounter
+
+aten = torch.ops.aten
+
+__all__ = [
+    "MemoryResidencySimulator",
+    "MemoryResidencyCounter",
+    "_access_convolution_backward",
```
AllenYolk, line 18: 💡 `_access_convolution_backward` is a private function (prefixed with `_`) but it's exported in `__all__`. Private helpers should not be in `__all__` -- this pollutes the public API. Remove it from `__all__` unless it's intentionally part of the public surface.

spikingjelly/activation_based/op_counter/memory_residency.py
```diff
@@ -0,0 +1,521 @@
...
+    aten.sum.default: _access_sum_mean,
+    aten.sum.dim_IntList: _access_sum_mean,
+    aten.mean.dim: _access_sum_mean,
+    aten.where.self: _access_where,
+    aten.where.ScalarOther: _access_where,
+    aten.where.ScalarSelf: _access_where,
+}
+
+
+class MemoryResidencySimulator:
```
AllenYolk, line 201: ⚠️ `MemoryResidencySimulator` has no docstring. Every other public class in the `op_counter` package has bilingual (CN/EN) docstrings per AGENTS.md. This module is 521 lines -- add docstrings explaining the three-tier cache model, eviction policy, and usage API.

spikingjelly/activation_based/op_counter/memory_residency.py
```diff
@@ -0,0 +1,521 @@
...
+        }
+
+    def get_move_bits_by_edge(self) -> dict[str, int]:
+        return dict(self.move_bits_by_edge)
+
+    def get_move_bits_by_op(self) -> dict[str, dict[str, int]]:
+        return {op: dict(d) for op, d in self.move_bits_by_op.items()}
+
+
+class MemoryResidencyCounter(BaseCounter):
```
AllenYolk, line 466: ⚠️ Same as above -- `MemoryResidencyCounter` needs a bilingual docstring explaining its purpose, the delta-based level accounting pattern, and how it differs from `MemoryResidencySimulator`.

AllenYolk:
## Code Review Summary

**Verdict: Request Changes** (2 issues need attention, 3 suggestions)

---

### 🔴 Critical

- **core.py:124** — All bilingual docstrings (CN/EN) stripped from public APIs: `NeuroMCRuntimeEnergyReport`, `NeuroMCEnergyProfiler`, `estimate_neuromc_runtime_energy`, `MemoryHierarchyConfig`. This violates the project's AGENTS.md documentation requirements. The new report structure has 10+ new fields (`energy_mac_pj`, `energy_by_core_type`, `energy_by_process_key`, `mapping_summary`, etc.) that are completely undocumented. **Must restore before merge.**

### ⚠️ Warnings

- **core.py:312** — `__exit__` does not clean up `_neuromc_last_input` attributes monkey-patched onto Conv/Linear modules during `bind_model()`. After profiling, modules retain stale tensor references — memory leak and attribute pollution on shared modules.
- **core.py:64** — `_AUXILIARY_ATEN_OPS` includes `aten.add.Tensor` / `aten.mul.Tensor` etc., which the old profiler explicitly counted as primitives with energy costs (0.548 pJ/add, 0.812 pJ/mul). These are now silently free. If intentional in the exact model, document it; if not, remove from the set.
- **Breaking API change** — Removed parameters from public functions: `op_cost_pj`, `memory_cost_pj_per_bit`, `memory_level_weights`, `memory_model` from `estimate_neuromc_runtime_energy()`; `memory_model` field from `MemoryHierarchyConfig`. Existing downstream code passing these will get `TypeError`. This needs a deprecation path or migration guide.

### 💡 Suggestions

- **memory_residency.py:68** — `_collect_tensors` (line 46) and `_extract_tensors` (line 68) are identical. Consolidate.
- **core.py** — `_is_spike_tensor` (core.py:125) duplicates `_is_spike` (utils.py:69). Consider importing from utils or consolidating.
- **core.py:1255** — `get_report()` lost its docstring. It now has complex behavior (hook vs trace fallback, `ValueError` on unsupported ops) that deserves documentation.
- **core.py** — The `_fallback_fragments_from_trace` method is ~120 lines of deeply nested if/elif. Consider extracting helper methods for the gemm backward and conv forward paths.

### ✅ Looks Good

- Clean architectural separation: memory residency counter promoted to top-level, NeuroMC focused on exact energy modeling.
- The fragment-based approach (module hooks → `_Fragment` dataclass → energy computation) is a significant improvement over the old dispatch-counter pattern.
- `base_counter.py` graceful handling of unknown ops (return 0) is a good fix.
- `add_counter.py` fix for `_bias_sizes` and the `isinstance(out, (tuple, list))` guard are correct.
- Test coverage is solid for the new exact model — spike detection, batch norm, training flow with bp_grad/wg/optimizer, weight reuse, frozen params, error paths.
- The `MemoryInstanceSpec` + `MemoryHierarchyConfig` redesign with hardware-specific instances is well-structured.

---
*Reviewed by Hermes Agent • [PR #676](https://github.com/fangwei123456/spikingjelly/pull/676)*

AllenYolk:
@coderabbitai please review the latest validated update on this PR. @gemini-code-assist please review the latest validated update as well.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -142,725 +194,1247 @@ def _resolve_loss_fn(loss_fn: Callable | None):
...
+                handle.remove()
+            self._hook_handles.clear()
+            if self._bound_model is not None:
+                for module in self._bound_model.modules():
+                    if hasattr(module, "_neuromc_last_input"):
+                        delattr(module, "_neuromc_last_input")
+            self._model_bound = False
+            self._bound_model = None
+        finally:
+            return self._trace_mode.__exit__(exc_type, exc, tb)
```
AllenYolk: 🔴 **Critical: `return` in `finally` silently suppresses all exceptions**

In Python, a `return` inside a `finally` block suppresses any pending exception, regardless of the return value. This means any exception raised inside `with profiler:` (e.g., from `bind_model`, forward hooks, or `record_optimizer_step`) will be silently eaten rather than propagated.

This would cause tests like `test_neuromc_exact_sgd_optimizer_is_rejected` and `test_neuromc_exact_ann_forward_input_is_rejected` to fail, since the `ValueError` they expect is raised inside the `with profiler:` block.

**Fix:** Move the `return` outside the `finally`:
```python
def __exit__(self, exc_type, exc, tb):
    self._active = False
    try:
        for handle in self._hook_handles:
            handle.remove()
        self._hook_handles.clear()
        if self._bound_model is not None:
            for module in self._bound_model.modules():
                if hasattr(module, "_neuromc_last_input"):
                    delattr(module, "_neuromc_last_input")
        self._model_bound = False
        self._bound_model = None
    finally:
        trace_ret = self._trace_mode.__exit__(exc_type, exc, tb)
    return trace_ret
```

spikingjelly/activation_based/op_counter/memory_residency.py
```diff
@@ -0,0 +1,528 @@
...
+        if key in self.reg_cache:
+            self._touch_inclusive_levels(key)
+            self._record_level_rw("reg", "write_bits", bits, op_name)
+            return
+
+        if key in self.sram_cache:
+            self._touch(self.sram_cache, key)
+            if self._insert_reg(key, bits, op_name):
+                self._record_level_rw("sram", "read_bits", bits, op_name)
+                self._record_move("sram", "reg", bits, op_name)
```
AllenYolk, line 434: ⚠️ **Double-counting of `reg write_bits` during SRAM→reg promotion**

`_record_move("sram", "reg", bits, op_name)` already records `reg write_bits += bits` internally (line 247 of memory_residency.py). The explicit `_record_level_rw("reg", "write_bits", ...)` on the next line adds `reg write_bits += bits` again.

The same double-count existed in the old `MemoryResidencySimulator` (pre-existing), so this is a carry-over rather than a regression. But it's worth noting — if the hardware model intends reg write_bits to count only the tensor write (not the cache management), one of these should likely be removed.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -114,24 +132,58 @@ class NeuroMCRuntimeEnergyReport:
...
+    input_numel: int
+    weight_numel: int
+    output_numel: int
+    mac_count: int
+    conv_type: str = "--"
+    b_type: int = 0
+    t_type: int = 0
+    source: str = "trace"
+
+def _tensor_numel(x: Any) -> int:
```
AllenYolk, line 250: 💡 Minor PEP 8: Two blank lines are expected before top-level definitions. There's only one blank line between `_Fragment` (ends line 175) and `_tensor_numel`.

AllenYolk:
## Code Review Summary

**Verdict: REQUEST_CHANGES** (1 critical issue, 2 observations)

### Architecture Assessment

This is a well-structured refactoring that replaces the generic dispatch-counter model with a module-hook-based exact profiler. The new `_Fragment`-centric approach maps directly to hardware compute/memory processes, which is a significant improvement for NeuroMC energy modeling accuracy. Key architectural wins:

- `bind_model` + forward/backward hooks replace per-op counters — cleaner separation
- `MemoryInstanceSpec` with 21 hardware presets replaces generic per-bit costs — matches the paper's exact model
- `suspend()` context manager cleanly excludes loss/optimizer.step() from profiling
- Legacy kwargs gracefully deprecated with `DeprecationWarning`
- Generic `MemoryResidencyCounter` promoted to `op_counter.memory_residency` — good reuse

### 🔴 Critical

- **`core.py:312`** — `return` in `finally` block silently suppresses all exceptions from `with profiler:` blocks. In Python, `return` inside `finally` always suppresses any pending exception regardless of the return value. This means errors raised inside `estimate_neuromc_runtime_energy` (e.g., SGD optimizer rejection at `record_optimizer_step`, ANN-like input rejection from forward hooks) would be silently swallowed instead of propagated. Tests like `test_neuromc_exact_sgd_optimizer_is_rejected` and `test_neuromc_exact_ann_forward_input_is_rejected` likely fail against this code. Fix: move the `return` outside the `finally` block (see inline comment for patch).

### ⚠️ Observations

- **`memory_residency.py:393-394`** — Double-counting of `reg write_bits` during SRAM→reg promotion. `_record_move("sram", "reg", ...)` already records `reg write_bits` internally, then the explicit `_record_level_rw("reg", "write_bits", ...)` adds it again. Pre-existing from old `MemoryResidencySimulator`, not a regression — but worth revisiting if the hardware model intent is to count only the tensor write.

- **Behavioral changes worth documenting**: (a) forward inputs must now be spike/binary (ANN-like inputs raise `ValueError`), (b) only Adam/AdamW optimizers supported, (c) `loss_fn` required when `optimizer` is provided, (d) `cmp` primitive renamed to `comp` in `primitive_counts`. These are breaking API changes for existing users — consider a migration note or `CHANGES.md` entry.

### ✅ Looks Good

- `_add_convolution_backward` fix (renaming `bias` to `_bias_sizes` and checking `out[2]`) correctly addresses a pre-existing bug where `args[3]` was misidentified as the bias tensor
- `base_counter.py` guard for unknown ops (`rule = self.rules.get(func); if rule is None: return 0`) prevents KeyError crashes
- `_is_spike` / `_spike_nnz` empty-tensor guards and `logical_or_` → `logical_or` fix are solid
- Config `validate()` strictness (only `neuromc_like_v1` preset, 32nm) prevents misuse
- Docstring type annotation fixes across `misc.py`, `rnn.py`, `datasets/base.py` are correct
- Test coverage is comprehensive — new tests exercise exact profiler behavior (SG rejection, BN breakdown, weight reuse, Conv3d rejection, etc.)

---
*Reviewed by Hermes Agent*

AllenYolk (commented):
See inline comments and summary above for details.

AllenYolk:
@coderabbitai please review the latest validated update on this PR. @gemini-code-assist please review the latest validated update as well.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -142,725 +194,1252 @@ def _resolve_loss_fn(loss_fn: Callable | None):
...
-        )
-        self._stage_memory_level_bits: dict[str, dict[str, int]] = defaultdict(
-            lambda: defaultdict(int)
+        self._suspended = False
+        self._model_bound = False
+        self._optimizer: torch.optim.Optimizer | None = None
+        self._hook_handles = []
+
+    def bind_model(self, model: nn.Module):
+        if self._model_bound:
```
AllenYolk, line 381: ⚠️ **Warning:** `bind_model` silently returns if already bound. If called with a different model, the second model's hooks are never registered — but no error is raised. Consider either raising `RuntimeError` or re-binding (removing old hooks first).

spikingjelly/activation_based/op_counter/neuromc/utils.py
```diff
@@ -69,15 +69,19 @@ def _is_spike(x: torch.Tensor | None) -> bool:
         return False
     if x.dtype == torch.bool:
         return True
-    return bool(x.eq(0).logical_or_(x.eq(1)).all().item())
+    if x.numel() == 0:
+        return False
+    return bool(x.eq(0).logical_or(x.eq(1)).all().item())
```
AllenYolk, line 74: ⚠️ **Warning:** `_is_spike` scans the **entire tensor** (`.eq(0).logical_or(x.eq(1)).all()`) and is called on every forward hook. For large activation tensors this is O(N) with 3 intermediate boolean tensors. Consider an early-exit approach like `x.min().item() >= 0 and x.max().item() <= 1` which short-circuits on the range check, or sampling.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -142,725 +194,1252 @@ def _resolve_loss_fn(loss_fn: Callable | None):
...
-        rw_delta = _diff_nested_dict(
-            after["memory_level_rw_bits"], before["memory_level_rw_bits"]
+        return _Fragment(
+            stage=stage,
+            phase="forward",
+            op_name="conv.forward",
+            core_type="fp_soma",
+            process_key="with_nothing",
+            loop_dims=loop_dims,
+            input_precision_bits=1 if _is_spike(x) else 16,
```
AllenYolk: 💡 **Suggestion:** `_is_spike(x)` was already validated at line 444, so this ternary will always evaluate to `1`. Consider using `1` directly (or asserting `_is_spike(x)`) to avoid dead code and reader confusion. Same pattern at line 507.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -114,24 +132,58 @@ class NeuroMCRuntimeEnergyReport:
...
+    return int(x.numel())
+
+
+def _module_channel_count(x: torch.Tensor) -> int:
+    if x.ndim >= 2:
+        return int(x.shape[1])
+    return int(x.numel())
 
 
 def _resolve_loss_fn(loss_fn: Callable | None):
```
AllenYolk, line 307: 💡 **Suggestion:** `_resolve_loss_fn` returns its input unchanged for both `nn.Module` and plain callables — it's a no-op. Consider removing it and calling `loss_fn` directly.

spikingjelly/activation_based/op_counter/neuromc/config.py
```diff
@@ -2,222 +2,154 @@
...
+            preset_name="neuromc_like_v1",
+            technology_nm=32,
+            level_order=("dram", "sram", "reg", "noc"),
+            memory_instances=dict(_MEMORY_INSTANCES),
+            zero_dram_in_paper_energy=True,
+            zero_noc_in_paper_energy=True,
+            zero_sram_high_directions=True,
         )
 
     def copy(self):
```
AllenYolk, line 118: 💡 **Suggestion:** `copy()` manually duplicates dataclass fields instead of using `copy.copy(self)` or `dataclasses.replace()`. If new fields are added to `MemoryHierarchyConfig`, `copy()` must be updated in sync — easy to forget.

spikingjelly/activation_based/op_counter/neuromc/memory_residency_counter.py
```diff
@@ -1,252 +1,9 @@
...
-from torch.overrides import resolve_name
-from torch.utils._pytree import tree_flatten
-
-from .base_counter import NeuroMCBaseCounter
-from .config import MemoryHierarchyConfig
-from .residency import MemoryResidencySimulator
-from .utils import _infer_stage, _tensor_bits
-
-aten = torch.ops.aten
+from ..memory_residency import (
```
AllenYolk: ⚠️ **API breakage:** The old `NeuroMCMemoryResidencyCounter.__init__` accepted `memory_config: MemoryHierarchyConfig` as a positional/keyword arg. The new `MemoryResidencyCounter` uses keyword-only `config: Any`. Anyone calling `NeuroMCMemoryResidencyCounter(memory_config=...)` will get `TypeError`. Consider adding a `memory_config` → `config` alias or a deprecation shim.

spikingjelly/activation_based/op_counter/neuromc/__init__.py
```diff
@@ -1,30 +1,13 @@
-from .config import MemoryHierarchyConfig
+from .config import MemoryHierarchyConfig, MemoryInstanceSpec
```
AllenYolk, line 1: ⚠️ **Breaking change:** `MemoryResidencySimulator` was removed from `__all__` exports. Users doing `from ...neuromc import MemoryResidencySimulator` will break. This class is the primary public API for the residency simulator — recommend re-exporting it.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -142,725 +194,1252 @@ def _resolve_loss_fn(loss_fn: Callable | None):
...
+                self._make_linear_backward_fragments(stage, module, grad_input, grad_output)
+            )
+        elif isinstance(module, (nn.BatchNorm1d, nn.BatchNorm2d, nn.BatchNorm3d)):
+            self._fragments.append(self._make_bn_backward_fragment(stage, grad_input, grad_output))
+        elif isinstance(module, BaseNode):
+            self._fragments.append(
+                self._make_soma_backward_fragment(stage, grad_input, grad_output)
+            )
 
+    def _make_loop_dims(
```
AllenYolk, line 565: 💡 **Suggestion:** `_make_loop_dims` takes 7 separate `int` parameters. A `LoopDims` dataclass or TypedDict would be cleaner and less error-prone than positional args.

AllenYolk:
## Code Review Summary

**Verdict: COMMENT** — Solid refactor with high-quality tests. A few items need attention before merge.

### ⚠️ Warnings

- **`neuromc/__init__.py`** — `MemoryResidencySimulator` removed from exports. This is the primary public API for the residency simulator; `from ...neuromc import MemoryResidencySimulator` will break for downstream users. Recommend re-exporting.
- **`neuromc/memory_residency_counter.py`** — Constructor signature changed: `memory_config` → `config` (keyword-only). Existing callers using `memory_config=...` will get `TypeError`. Add a deprecation alias.
- **`utils.py:74`** — `_is_spike` scans the entire tensor on every forward hook. O(N) with 3 intermediate boolean tensors. Consider range-based early exit or caching.

### 💡 Suggestions

- **`core.py:263`** — `bind_model` silently ignores second call with different model. Consider raising or re-binding.
- **`core.py:420`** — `_make_loop_dims` takes 7 positional ints. A `LoopDims` dataclass would be cleaner.
- **`core.py:189`** — `_resolve_loss_fn` is a no-op; can be inlined.
- **`core.py:471,507`** — `_is_spike(x)` ternary is dead code after earlier validation.
- **`config.py:111`** — Manual `copy()` instead of `dataclasses.replace()`; fragile if fields are added.

### ✅ Looks Good

- Clean dataclass design for `NeuroMCRuntimeEnergyReport`, `_Fragment`, `_TraceEvent`
- `MemoryHierarchyConfig` frozen dataclass prevents accidental mutation
- `suspend()` context manager is well-tested and correctly implemented
- Test coverage is thorough: linear reports, SG/BN breakdowns, training stages, weight reuse, SGD rejection, Conv3d rejection, frozen params, error conditions
- Deprecation warnings for legacy kwargs are user-friendly
- `_record_move` correctly separates dst-write from caller-managed src-read — cleaner than the old double-counting
- `_tensor_key` semantic change (storage-level identity) is correct for memory hierarchy simulation
- The extraction from `neuromc/residency.py` → `memory_residency.py` is clean

---
*Reviewed by Hermes Agent*

AllenYolk:
@coderabbitai @gemini-code-assist

Validated follow-up fixes pushed in `c0c2e8dd`:
- avoid retaining live tensors in NeuroMC trace events
- merge functional GEMM trace fragments with hook fragments without double-counting
- fix optimizer exact `with_opt` scalar accounting
- pre-scan `Conv3d` before hook registration and reject rebinding to a different model
- restore `MemoryResidencySimulator` export and `memory_config` compatibility shim
- handle multi-step `[T, N, C, ...]` hook layouts correctly

Validation:
- `pytest test/activation_based/test_op_counter_neuromc_runtime.py -q -k "mixed_hook_and_functional_mm or optimizer_counts_mixed_parameter_shapes or trace_events_do_not_retain_live_tensors or bind_model_rejects_conv3d_before_registering_hooks or bind_model_rejects_rebinding_different_model or legacy_memory_config_alias or repeated_forward_reuses_weights or b_type_reuse_hint"`
- `pytest test/activation_based/test_op_counter_neuromc_runtime.py -q`

AllenYolk (commented):
## Hermes Agent Review

**Verdict: COMMENT** (observations and suggestions — no blocking issues)

This is a well-structured refactoring that cleanly separates memory-residency simulation from NeuroMC-specific energy estimation. The shift from aten-op dispatch counters to module-based hooks with fragment-based energy modeling is a significant architectural improvement.

---

### ⚠️ Warnings

1. **`_record_move` no longer records source-level reads** (`memory_residency.py:306-311`)
   The new `_record_move` only records `write_bits` on the destination. Source-level `read_bits` is now tracked by explicit caller-side `_record_level_rw` calls. The net effect is equivalent for all current call sites, but the method name is misleading — it describes a "move" but only writes the destination. Consider renaming to `_record_transfer` or adding a docstring clarifying that callers are responsible for source reads.

2. **Duplicated `_infer_stage` logic** (`memory_residency.py:115-129` vs `neuromc/utils.py`)
   `_infer_stage` is defined in both `memory_residency.py` and `neuromc/utils.py` with identical logic. If the stage inference logic diverges in the future, this will silently cause bugs. Recommend importing from one canonical location.

3. **`NeuroMCRuntimeEnergyReport` field count doubled** (8 → 17 fields)
   The report dataclass now has 17 fields. Consider whether `counts_by_core_type`, `counts_by_process_key`, and `mapping_summary` belong in the report or should be accessed via separate methods on the profiler. The report is a return value from a convenience function, so keeping it lean improves API usability.

---

### 💡 Suggestions

4. **`_insert_sram` and `_insert_reg` have near-identical eviction loops** (`memory_residency.py:323-375` and `377-422`)
   Both methods contain the same LRU eviction pattern with minor variations (reg→sram vs sram→dram). Extracting a `_evict_until_fit(cache, capacity_key, target_bits, op_name, next_level)` helper would reduce the ~100 lines of duplicated logic.

5. **`neuromc_like_v1()` classmethod is redundant** (`config.py`)
   All arguments passed in the classmethod match the dataclass field defaults exactly. The method could be simplified to `return cls()` or removed entirely. If it exists for explicit construction clarity, add a comment explaining why.

6. **`_optimizer_fragment` repurposes conv loop dims for optimizer params** (`core.py:3063-3075`)
   `K` is set to 1D-param count, `FX` to multi-D-param count. These don't correspond to actual conv loop dimensions. A brief comment would help future readers understand the mapping.

7. **Deprecated kwargs in `estimate_neuromc_runtime_energy`** (`core.py:3456`)
   The function signature still accepts `op_cost_pj`, `memory_cost_pj_per_bit`, `memory_level_weights`, `memory_model` — all of which are now silently ignored with a deprecation warning. Consider adding a `# TODO: remove in v2.x` comment or opening an issue to track removal.

8. **`_AUXILIARY_ATEN_OPS` contains arithmetic ops also in `_RESIDENCY_ACCESS_RULES`** (`core.py:1148-1209`)
   `aten.add.Tensor`, `aten.mul.Tensor`, etc. appear in both `_AUXILIARY_ATEN_OPS` (for unsupported-op filtering) and `_RESIDENCY_ACCESS_RULES` (for memory tracking). Since the profiler no longer uses the memory counter path, the access rules are effectively dead code in the context of `NeuroMCEnergyProfiler`. If `MemoryResidencyCounter` is still public API, this is fine — just noting the dual role.

---

### ✅ Looks Good

- **`_is_spike` empty tensor fix** (`utils.py:68-71`): `logical_or_` → `logical_or` prevents in-place mutation of comparison results. `numel() == 0` guard is correct.
- **`_trace_events` uses lightweight snapshots**: `_snapshot_trace_value` creates `_TraceTensor` dataclass instances instead of holding live tensors — avoids memory leaks.
- **`bind_model` validates Conv3d before registering hooks**: On failure, no hooks are left behind. Clean.
- **`base_counter.py` graceful unknown-op handling**: Returns 0 instead of KeyError — prevents crashes when unregistered ops are dispatched.
- **`_fragment_signature` deduplication**: Module-sourced and trace-sourced fragments sharing the same signature are correctly deduplicated in `get_report()`.
- **Test coverage**: 25+ test cases covering exact energy decomposition, error paths, multi-step mode, optimizer accounting, frozen params, Conv3d rejection, and edge cases. The hardcoded expected values in `test_neuromc_exact_optimizer_counts_mixed_parameter_shapes` are mathematically verified.
- **Backward compatibility**: `MemoryResidencySimulator` re-exported from `neuromc` submodule; `NeuroMCMemoryResidencyCounter` wraps new `MemoryResidencyCounter` with deprecation shim.
- **Docstring type-annotation fixes** in `misc.py`, `rnn.py`, `datasets/base.py` are correct (e.g., `:type threshold0: float` → `:type eval_threshold: float`).

---
*Reviewed by Hermes Agent*

AllenYolk:
## Review Summary

**Verdict: COMMENT** — no blocking issues found.

This refactoring cleanly separates memory-residency simulation into a generic module and rewrites the NeuroMC energy profiler around module hooks + fragment-based energy estimation. The architecture shift is well-motivated and correctly implemented.

**Key items to consider:**
1. `_record_move` semantic change (source reads moved to callers) — equivalent behavior but the method name is now misleading
2. Duplicated `_infer_stage` between `memory_residency.py` and `neuromc/utils.py`
3. `_insert_sram`/`_insert_reg` eviction loops could share a helper (~100 lines of duplication)
4. Report dataclass field count doubled (8→17) — consider what belongs in the report vs. profiler methods

All tests pass the math verification. Good test coverage for the new fragment-based model.

---
*Reviewed by Hermes Agent*

AllenYolk:
@coderabbitai @gemini-code-assist

Validated follow-up fixes pushed in `ef83ef7e`:
- restore `NeuroMCMemoryResidencyCounter` root re-export and positional `memory_config` compatibility
- keep `MemoryHierarchyConfig.neuromc_like_v1(memory_model=...)` backward-compatible with deprecation warning
- include op/source in fragment signatures while keeping module-vs-trace overlap dedupe normalized on `B*T`
- allow supplemental functional `addmm`/conv/BN fallback paths, but fail fast on unsupported multi-step/3D trace fallback cases
- skip optimizer params with `grad is None` when building exact `with_opt` fragments

Validation:
- `pytest test/activation_based/test_op_counter_neuromc_runtime.py -q -k "mixed_hook_and_functional_addmm or optimizer_skips_params_without_grad or legacy_memory_config_positional_alias or neuromc_like_v1_legacy_memory_model_is_ignored or multi_step_trace_batchnorm_is_rejected or memory_config_api"`
- `pytest test/activation_based/test_op_counter_neuromc_runtime.py -q`

Skipped intentionally:
- `core_type` no-op review item: valid concern, but not a minimal bug fix; changing it now would require redesigning the public exact-runtime API and fragment selection semantics.

spikingjelly/activation_based/op_counter/memory_residency.py
```diff
@@ -0,0 +1,569 @@
...
+        self.op_level_rw_bits[op_name][level][rw] += bits
+        self.op_level_rw_bits[op_name][level]["total_bits"] += bits
+
+    def _record_move(self, src: str, dst: str, bits: int, op_name: str):
+        if bits <= 0:
+            return
+        edge = f"{src}->{dst}"
+        self.move_bits_by_edge[edge] += bits
+        self.move_bits_by_op[op_name][edge] += bits
+        self._record_level_rw(dst, "write_bits", bits, op_name)
```
AllenYolk, line 248: 💡 **Note:** `_record_move` now only records destination `write_bits`; the old version also recorded source `read_bits`.

This is actually a **bug fix** — the old code double-counted reads during data movement (once via `_record_move`, once via explicit `_record_level_rw` calls in `on_tensor_read`/`on_tensor_write`). The new approach is cleaner: source reads are recorded explicitly at call sites, and `_record_move` only handles the destination write.

However, this changes the observable `level_rw_bits` for direct users of `MemoryResidencySimulator` who were relying on the old totals. Worth noting in release notes if anyone is comparing against pre-refactor baselines.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -142,725 +312,1454 @@ def _resolve_loss_fn(loss_fn: Callable | None):
...
+        elif fragment.process_key == "with_bn":
+            if fragment.phase == "forward":
+                counts["add"] += oyoxkbt * 3 + kt * 2
+                counts["mul"] += oyoxkbt * 3 + kt * 4
+                counts["sqrt"] += kt
+            else:
+                counts["add"] += oyoxcbt * 7
+                counts["mul"] += oyoxcbt * 3 + ct * 22
+                if fragment.conv_type == "without_bp_bn":
+                    counts["add"] = 0
```
AllenYolk, line 1269: ⚠️ **Suggestion:** This uses `=` (full assignment) rather than subtraction. It works correctly today because `with_bn` backward fragments only accumulate BN-related primitives. But if this `elif` branch ever gains contributions from another `process_key`, the full zero-out would silently drop them.

Consider either:
- Adding a comment like `# BN backward has no add/mul when conv_type omits BP BN`
- Or resetting only the BN-specific delta (e.g., tracking the pre-computation counts)

Not blocking — just a fragility note for future extension.

spikingjelly/activation_based/op_counter/neuromc/core.py
```diff
@@ -142,725 +312,1454 @@ def _resolve_loss_fn(loss_fn: Callable | None):
...
+    def _gemm_backward_fragment_kind(
+        self, x: torch.Tensor | _TraceTensor, y: torch.Tensor | _TraceTensor, out: Any
+    ) -> str | None:
+        x_req = bool(x.requires_grad)
+        y_req = bool(y.requires_grad)
+        if x_req and (not y_req):
+            return "wg"
+        if (not x_req) and y_req:
+            return "bp_grad"
+        if self._matches_trainable_param_shape(out):
```
AllenYolk, line 1001: 💡 **Suggestion:** The fallback `_matches_trainable_param_shape(out)` heuristic (line 1001) could misclassify when multiple parameters share the same shape. For example, two `nn.Linear(8, 8)` layers would both match the same output shape.

In practice this only triggers in the trace fallback path (not the hook path), and the `supplemental_only` mode filters most of these out. But for mixed models where both hook-captured and trace-captured GEMM backward ops coexist, consider storing the parameter reference in `_TraceEvent` to disambiguate more reliably.

AllenYolk:
## Code Review Summary

**Verdict: COMMENT** — No blocking issues. Clean refactor with solid architecture.

---

### ✅ Looks Good

- **Clean architectural separation.** Promoting `MemoryResidencySimulator` to `op_counter/memory_residency.py` as a generic component, while keeping `neuromc/` focused on exact energy profiling, is the right layering. The old monolithic dispatch-counter approach is well-replaced by the hook + trace-event architecture.

- **Bug fixes are correct.**
  - `logical_or_` → `logical_or` in `_is_spike` / `_spike_nnz`: prevents in-place mutation of input tensors. Good catch.
  - `_bias_sizes` rename in `_add_convolution_backward`: the 4th arg of `aten.convolution_backward` is `bias_sizes` (a list), not the bias tensor itself.
  - Empty tensor guard in `_is_spike` / `_spike_nnz`: avoids calling `.all()` on empty tensors.
  - `_record_move` simplification actually fixes old double-counting of reads during data movement.

- **Backward compatibility handled well.** `NeuroMCMemoryResidencyCounter` provides a deprecation shim, `neuromc/residency.py` re-exports from the new location, and `estimate_neuromc_runtime_energy` warns on legacy kwargs.

- **Test suite is comprehensive.** 30+ tests covering: exact profiler fields, SG/bn/optimizer breakdown, weight reuse, Conv3d rejection, ANN-input rejection, empty tensors, rebinding, legacy aliases, multi-step layout. Good coverage of edge cases.

- **`_TraceTensor` snapshots** prevent live tensor retention in trace events — important for memory safety during long profiling runs.

- **Optimizer integration** is clean: `bind_optimizer` + `record_optimizer_step` + `suspend()` gives fine-grained control without coupling to the profiler's dispatch mode.

### ⚠️ Warnings

1. **`_record_move` behavioral change** (memory_residency.py:242-248) — See inline comment. The new version only records destination writes; old code recorded both source reads and destination writes. This is a correctness fix (old code double-counted reads) but changes observable `level_rw_bits` for direct `MemoryResidencySimulator` users. Mention in release notes.

2. **`memory_traffic_counter.py` deleted without migration path** — Users importing `NeuroMCMemoryTrafficCounter` will get `ImportError`. The weighted-approximation model is gone entirely. If any downstream code uses it, there's no deprecation shim (unlike `NeuroMCMemoryResidencyCounter` which has one). This is fine if the weighted model was purely internal, but worth confirming.

### 💡 Suggestions

1. **`without_bp_bn` zeroing** (core.py:1268-1270) — See inline comment. Using `=` instead of subtraction is fragile if the branch gains additional contributions. A short comment would help future readers.

2. **GEMM backward shape heuristic** (core.py:1001-1003) — See inline comment. `_matches_trainable_param_shape(out)` could misclassify when multiple parameters share the same shape in the trace fallback path. Consider storing parameter refs in `_TraceEvent` for more reliable disambiguation.

3. **Internal method testing** — `test_neuromc_exact_optimizer_counts_mixed_parameter_shapes` directly calls `_optimizer_fragment`, `_extra_counts`, `_extra_memory_for_fragment`. These are implementation details. The numerical assertions are valuable, but consider extracting the tested logic into a public `compute_optimizer_energy()` method if you want to keep this level of testing.

---

*Reviewed by Hermes Agent*

PR #695, opened 2026-06-27
Add local threshold balancing ANN2SNN recipe

AllenYolk:
## Summary

Add a training-free ANN2SNN recipe based on local threshold balancing and update the ANN2SNN package organization/docs around recipe-based conversion.

Main changes:
- Split `spikingjelly.activation_based.ann2snn.recipes` into a `recipes/` package.
- Add `LocalThresholdBalancingRecipe` with channel-wise threshold scaling.
- Add `ChannelVoltageScaler` and `estimate_delay_start` for ANN2SNN delayed-readout evaluation.
- Move/add `HalfThresholdIFNode` as an activation-based neuron.
- Extend MNIST and ResNet-18 ImageNet examples and tutorial comparison tables.
- Add tests for LTB conversion, channel-wise scaling, half-threshold neuron behavior, delay estimation, and ANN state preservation.

## Accuracy Results

MNIST full test, `T=32`:
- ANN top-1: 98.70
- RobustNorm legacy scalar top-1: 98.35
- LocalThresholdBalancingRecipe top-1: 98.53

ImageNet ResNet-18 full val, `T=32`, calibration subset 50000, `delay_start=auto`:
- ANN top-1/top-5: 69.76 / 89.08
- RobustNorm legacy scalar top-1/top-5: 12.57 / 28.99
- LocalThresholdBalancingRecipe top-1/top-5: 65.45 / 86.60

The tutorial explicitly notes that this is a SpikingJelly recipe implementation of the main local-threshold-balancing idea, not a full reproduction of the original paper pipeline.

Reference checked against arXiv official metadata:
- Bu T, Li M, Yu Z. Inference-Scale Complexity in ANN-SNN Conversion for High-Performance and Low-Power Applications. arXiv:2409.03368, 2024. Accepted by CVPR 2025.

## Validation

```bash
uv run python -m pytest test/activation_based/test_ann2snn.py test/activation_based/test_ann2snn_transformer.py -q
PYTHONPATH=$PWD uv run python -m sphinx -b html docs/source /tmp/spikingjelly-docs-final-ann2snn-check -W --keep-going
git diff --check
```

Result:
- `183 passed`
- Sphinx build succeeded
- `git diff --check` passed

## Summary by CodeRabbit

* **New Features**
  * Added new ANN2SNN conversion recipes, including local threshold balancing, channel-wise voltage scaling, delayed readout start estimation, and transformer spike-equivalent conversion.
  * Introduced `HalfThresholdIFNode`, expanded the public ANN2SNN API with `ChannelVoltageScaler` and delay-start estimation utilities.
  * Added CLI evaluation workflows for MNIST and ImageNet with timestep/recipe sweeps and optional JSON outputs.
* **Documentation**
  * Refreshed the ANN2SNN tutorial to match the current Recipe API and updated theory/examples.
* **Bug Fixes**
  * Improved checkpoint/download resilience and ensured conversion restores model train/eval state consistently.
* **Tests**
  * Expanded coverage for new conversions, delay estimation, voltage/threshold behavior, and download edge cases.


AllenYolk:
## Code Review — PR #695: Add local threshold balancing ANN2SNN recipe

**Verdict: Approve** (0 critical, 0 warnings, 2 suggestions)

Reviewed the full diff (+3818/-1345 across 16 files). Overall this is a well-structured PR with clean separation of concerns, comprehensive tests, and proper bilingual docstrings.

---

### ✅ Looks Good

- **Package restructuring**: `recipes.py` → `recipes/` package with clean re-exports in `__init__.py`. Backward-compatible — `from spikingjelly.activation_based.ann2snn.recipes import RateCodingRecipe` still works.
- **LTB recipe**: `LocalThresholdBalancingHook` implements per-batch threshold balancing with overflow-based convergence. The `replace_with_neurons` method correctly handles the ReLU→MaxPool2d fusion case.
- **HalfThresholdIFNode**: Correctly initializes membrane at `v_threshold / 2` via `set_reset_value("v", half_threshold)`. The `v_float_to_tensor` override properly re-initializes to half-threshold on shape change.
- **`MemoryModule.get_reset_value`**: Clean addition, counterpart to existing `set_reset_value`. Non-breaking.
- **`_safe_quantile`**: Good memory-bounded approximation for large activation tensors. The stride-based subsampling avoids OOM during calibration.
- **`estimate_delay_start`**: Proper try/finally cleanup of hooks. The analog passthrough in the probe hook (`torch.clamp(x, min=0, max=max_value)`) correctly simulates neuron behavior for downstream layers.
- **`ChannelVoltageScaler`**: Clean channel-wise scaling with proper validation. `_view_scale` correctly broadcasts 1D scale tensors.
- **Test coverage**: HalfThresholdIFNode (init, shape change, surrogate backward), LTB conversion (basic, pre-spike maxpool, validation errors), delay estimation (half-threshold formula, IF vs half-threshold ratio, multi-batch, hook cleanup, exception cleanup, unmatched graphs), channel-wise rate coding.
- **Docs**: CN/EN tutorials updated with new comparison tables. Code examples are consistent between languages. The `.. note::` about LTB being an engineering implementation (not full paper reproduction) is honest and appropriate.
- **rate_coding.py**: `channel_wise` and `half_threshold` options integrate cleanly with proper mutual-exclusion validation (channel_wise + custom rules/threshold_optimizer/neuron_factory raises ValueError).

### 💡 Suggestions

1. **`threshold_candidates` parameter accepted but unused** — `LocalThresholdBalancingRecipe.__init__` accepts `threshold_candidates` (default `(0.5, 0.75, 1.0, 1.25, 1.5)`), but the current LTB implementation uses per-batch overflow updates, not candidate selection. The docstring says "retained for compatibility with earlier experimental APIs", but users may still be confused about whether this parameter affects behavior. Consider either (a) removing it if there are no backward-compatibility consumers, or (b) adding a note in the docstring that it has no effect on the current implementation.

2. **Delay estimation math verification** — The `test_half_threshold_delay_matches_legacy_ltb_formula` test asserts `delay_start == 2` for a single-stage `Scaler(4.0) -> HalfThresholdIFNode(v_threshold=1.0) -> Scaler(4.0)` with `x=ones(2,2)`. Tracing through `_compute_delay_ratio`: `original_required_charge.mean() = 2.0`, `max_mean_activation = 16.0`, ratio = 0.125, `ceil(0.125) = 1`. This doesn't match the expected 2. Similarly `test_if_node_delay_is_twice_half_threshold_delay` expects IF delay = 4 but the same formula gives 1. The test passes in CI (183 passed), so the formula is likely correct and I'm miscounting the paths or the hook firing semantics — but worth a second look to make sure the delay estimation formula matches the documented intent.

---

*Reviewed by Hermes Agent*

AllenYolk:
## Supplementary Review — Additional findings from deep analysis

### ⚠️ Warning

1. **`rate_coding.py:335-339` — `pre_spike_maxpool=True` erases ReLU without replacing its uses.** When `pre_spike_maxpool=True`, the code skips `activation_node.replace_all_uses_with(node2)` and directly erases the ReLU node. If the ReLU has consumers other than MaxPool2d (e.g., a skip connection), those references will be broken. Current tests only use `nn.Sequential` models where ReLU has a single consumer, so this path isn't exercised with branching topologies. Consider adding a guard or documenting the single-consumer assumption.

2. **`delay.py` vs `rate_coding.py` — `_extract_batch_input` inconsistency.** `delay.py:27-35` only handles tensor or tuple/list[0], while `rate_coding.py:577-620` handles dicts and nested structures. Users with dict-based dataloaders will get a `TypeError` from `estimate_delay_start` but not from the recipe calibration. Consider aligning the two implementations.

### 💡 Additional Suggestions

3. **`integrate_and_fire.py` — `v_float_to_tensor` hardcodes `half_threshold`.** When the input shape changes (line ~676), the membrane is reinitialized to `self.v_threshold / 2.0` rather than `self.get_reset_value("v")`. If someone subclasses and overrides the reset value, the reinitialization would ignore the custom value. Low severity in practice.

4. **`delay.py:226` — Hook return value replaces neuron input with clamped value.** The probe hook returns `torch.clamp(x, min=0, max=v_threshold)`, which means the neuron's actual forward computation is bypassed during estimation. This is acceptable for the estimation heuristic but worth a docstring note.

---
*Reviewed by Hermes Agent — deep analysis pass*

AllenYolk:
Review feedback triage and update pushed in `f27ac37c`.

Implemented minimal fixes:
- Fixed the English tutorial typo from `__()` to `__call__()`.
- Restored the original ANN `.training` flag after `Converter.convert()`, with a regression test covering both train/eval mode.
- Fixed delayed-readout clamping to use `time_steps - _MIN_READOUT_STEPS`.
- Made delay estimation tolerate custom neuron classes without `get_reset_value()` by falling back on `v_reset`.
- Added `.contiguous()` before `_safe_quantile(...).kthvalue()` on strided/subsampled tensors.
- Rejected empty `ChannelVoltageScaler` scales and added test coverage.
- Added bounded `requests.get(..., timeout=30)` in ann2snn download paths.
- Ensured the MNIST checkpoint parent directory exists before download.
- Wrote ImageNet progress JSON atomically via temp file + replace.
- Made the ImageNet example default to CPU when CUDA is unavailable.
- Fixed `MemoryModule.get_reset_value()` to validate against registered memories, not just reset-value storage.

Intentionally skipped / not changed:
- Dictionary dataloader support in `estimate_delay_start`: this helper is intentionally scoped to the simple ANN2SNN evaluation path (tensor or first tuple/list item). More complex dataloader structures should be handled by user recipes or wrappers rather than expanding this helper's API surface.
- Branching-neuron delay estimation by accepting multiple post-scalers: current delayed-readout estimation is intentionally limited to the explicit scaler -> neuron -> scaler pattern produced by the supported recipes. For branched/custom topologies, silently choosing the first scaler would be ambiguous and could produce misleading delay estimates.
- `pre_spike_maxpool=True` branching support: current recipe support is for the sequential ReLU -> MaxPool pattern used by the supported CNN path. Multi-consumer activation graphs should be handled by a custom rule/recipe rather than implicitly rewriting ambiguous branches.
- `threshold_candidates` behavior: kept as documented compatibility metadata for earlier experiments; current LTB implementation uses the per-batch threshold-balancing update and does not claim candidate search.
- Delay formula suggestion from Hermes: rechecked through the existing tests; the current formula and hook behavior are covered by regression tests, so no formula change was made beyond the readout clamp fix above.

Validation run:

```bash
uv run python -m pytest test/activation_based/test_ann2snn.py::TestChannelVoltageScaler test/activation_based/test_ann2snn.py::TestLocalThresholdBalancingRecipe::test_recipe_does_not_mutate_ann_state_dict test/activation_based/test_ann2snn.py::TestDelayedReadoutEstimation -q
uv run python -m pytest test/activation_based/test_ann2snn.py test/activation_based/test_ann2snn_transformer.py -q
PYTHONPATH=$PWD uv run python -m sphinx -b html docs/source /tmp/spikingjelly-docs-pr-loop-check -W --keep-going
git diff --check
```

Results:
- Targeted tests: `14 passed`
- Full ann2snn tests: `184 passed`
- Sphinx build succeeded
- `git diff --check` passed

@coderabbitai review
@gemini-code-assist review

AllenYolk:
Addressed the latest actionable review-bot findings in `ba0aab1861d46b1db38321f4ada4baa3a1c17fc9`.

Implemented fixes:
- Preserve per-module training modes in `Converter.convert()` instead of only restoring the root module mode. This keeps mixed train/eval submodule states intact after conversion.
- Preserve model device and per-module training modes around `estimate_delay_start()`, since the helper temporarily moves the model to the target device and switches it to eval mode.
- Guard delayed-readout estimation against non-finite delay ratios before applying `ceil()`, so degenerate calibration probes do not crash the helper.
- Harden `download_url()` resume behavior: if a server ignores a resume `Range` request or returns an invalid `Content-Range`, restart from byte 0 and overwrite instead of appending a duplicate body.
- Added regression tests for the state restoration paths, non-finite delay handling, and unsafe resume fallback.

Validation run:
- `uv run python -m pytest test/activation_based/test_ann2snn.py::TestLocalThresholdBalancingRecipe::test_recipe_does_not_mutate_ann_state_dict test/activation_based/test_ann2snn.py::TestDelayedReadoutEstimation test/activation_based/test_ann2snn.py::TestDownloadUrl -q` -> 12 passed
- `uv run python -m pytest test/activation_based/test_ann2snn.py test/activation_based/test_ann2snn_transformer.py -q` -> 187 passed
- `PYTHONPATH=$PWD uv run python -m sphinx -b html docs/source /tmp/spikingjelly-docs-pr-loop-check -W --keep-going` -> passed
- `git diff --check` -> passed

Intentionally kept scope unchanged:
- I did not broaden `estimate_delay_start()` to recursively support dictionary dataloader batches. The current helper covers the simple ANN2SNN example path; custom batch structures should be handled by user recipes/wrappers.
- I did not broaden delayed-readout graph matching to ambiguous branching topologies by selecting a first post-scaler. Ambiguous topology support should be designed explicitly rather than guessed inside this helper.

@coderabbitai please review
@gemini-code-assist please review

AllenYolk:
Addressed the latest actionable low-risk review findings in `5a7964146ee9b8b9a9379a8707036ba0548f573b`.

Implemented fixes:
- Restored saved train/eval flags by assigning `module.training` directly instead of calling `module.train(...)` on every submodule, avoiding redundant recursive updates.
- Moved `estimate_delay_start()`'s temporary `model.eval().to(device)` transition inside the cleanup scope, so cleanup still runs if the device transition raises.
- Removed the redundant `int()` around `math.ceil(delay)`.
- Made `download_url()` close streamed responses, including retry/fallback responses, and made `Content-Range` unit parsing case-insensitive.
- Allowed `HalfThresholdIFNode(v_threshold=...)` to accept scalar PyTorch tensors, while rejecting non-scalar tensors with a clear `ValueError`.
- Added regression tests for scalar tensor thresholds and case-insensitive resume headers.

Validation run:
- `uv run python -m pytest test/activation_based/test_ann2snn.py::TestLocalThresholdBalancingRecipe::test_half_threshold_if_node_accepts_integer_threshold test/activation_based/test_ann2snn.py::TestLocalThresholdBalancingRecipe::test_half_threshold_if_node_accepts_scalar_tensor_threshold test/activation_based/test_ann2snn.py::TestLocalThresholdBalancingRecipe::test_half_threshold_if_node_rejects_non_scalar_tensor_threshold test/activation_based/test_ann2snn.py::TestDelayedReadoutEstimation test/activation_based/test_ann2snn.py::TestDownloadUrl -q` -> 14 passed
- `uv run python -m pytest test/activation_based/test_ann2snn.py test/activation_based/test_ann2snn_transformer.py -q` -> 190 passed
- `PYTHONPATH=$PWD uv run python -m sphinx -b html docs/source /tmp/spikingjelly-docs-pr-loop-check -W --keep-going` -> passed
- `git diff --check` -> passed

Intentionally not changed in this batch:
- I did not change the delayed-readout ratio formula to ignore the post-scaler. The current implementation follows the previously planned original-activation-space estimate and is covered by existing regression expectations. The suggested formula would change `delay_start` values and therefore conversion accuracy/benchmark results; that needs a separate accuracy validation pass rather than a small review-loop cleanup.
- Existing scope decisions remain unchanged: no recursive dictionary dataloader support in `estimate_delay_start()`, and no ambiguous branching-topology support by choosing the first post-scaler.

@coderabbitai please review
@gemini-code-assist please review

AllenYolk:
Addressed the latest actionable download-helper findings in `6381b06ab8a1b9a74b847ee6693d289a40d0b9a3`.

Implemented fixes:
- Updated `download_url()` progress accounting to use the actual received `len(chunk)` instead of a hardcoded chunk size.
- Tightened resumed-download validation: a resume response must be `206` and its `Content-Range` must match the requested start, end, and total size before appending.
- Tightened restart validation: restarted downloads accept either a full `200` response or a valid `206 bytes 0-(size-1)/size`; invalid restarted ranged responses now fail instead of writing corrupt data.
- Added a regression test for mismatched `Content-Range` end bytes forcing a clean restart.

Validation run:
- `uv run python -m pytest test/activation_based/test_ann2snn.py::TestDownloadUrl -q` -> 3 passed
- `uv run python -m pytest test/activation_based/test_ann2snn.py test/activation_based/test_ann2snn_transformer.py -q` -> 191 passed
- `PYTHONPATH=$PWD uv run python -m sphinx -b html docs/source /tmp/spikingjelly-docs-pr-loop-check -W --keep-going` -> passed
- `git diff --check` -> passed

Intentionally not changed:
- I again did not broaden `estimate_delay_start()` to reuse `RateCodingRecipe._extract_batch_input()` or support dictionary dataloaders. That remains outside this helper's intended minimal API boundary; users with custom batch structures should adapt their recipe/wrapper before calling delay estimation.

@coderabbitai please review
@gemini-code-assist please review

AllenYolk:
Addressed the latest actionable download-validation finding in `01b9a03f089223e87153d3ad1cfb0be762ee7c40`.

Implemented fixes:
- Centralized `download_url()` response validation in an internal helper.
- Validate the response status before opening/writing `dst`, so error payloads are not saved to disk.
- Preserve the stricter ranged-download rules from the previous commit: append only on a valid `206 Content-Range`, restart only with a full `200` or valid `206` from byte 0.
- Added a regression test that an HTTP error response raises before writing the destination file.

Validation run:
- `uv run python -m pytest test/activation_based/test_ann2snn.py::TestDownloadUrl -q` -> 4 passed
- `uv run python -m pytest test/activation_based/test_ann2snn.py test/activation_based/test_ann2snn_transformer.py -q` -> 192 passed
- `PYTHONPATH=$PWD uv run python -m sphinx -b html docs/source /tmp/spikingjelly-docs-pr-loop-check -W --keep-going` -> passed
- `git diff --check` -> passed

Intentionally not changed:
- I did not switch `module.training = training` back to `module.train(training)`. The direct assignment is deliberate here: it restores a captured per-module flag map in O(N), while calling `train()` on every module recursively is O(N^2) and can repeatedly perturb descendants while restoring mixed train/eval state. This code path is cleanup after temporary helper/converter mutations, not a public mode transition API.
- I did not broaden `estimate_delay_start()` to support dictionary dataloaders. That remains user recipe/wrapper responsibility for this helper's intentionally narrow API.

@coderabbitai please review
@gemini-code-assist please review

AllenYolk:
Addressed the latest actionable download robustness finding in `dfd3923b3dcd64302609d1e9663e86f87e77f377`.

Implemented fixes:
- The initial `download_url()` probe now calls `raise_for_status()` before reading headers, so HTTP errors fail before writing anything.
- Missing `content-length` is now handled by a single-pass non-resumable download path instead of raising `KeyError`.
- Added regression tests for probe HTTP errors and no-`content-length` downloads.

Validation run:
- `uv run python -m pytest test/activation_based/test_ann2snn.py::TestDownloadUrl -q` -> 6 passed
- `uv run python -m pytest test/activation_based/test_ann2snn.py test/activation_based/test_ann2snn_transformer.py -q` -> 194 passed
- `PYTHONPATH=$PWD uv run python -m sphinx -b html docs/source /tmp/spikingjelly-docs-pr-loop-check -W --keep-going` -> passed
- `git diff --check` -> passed

Intentionally not changed:
- `estimate_delay_start()` still does not support dictionary dataloader batches; custom batch extraction remains a user recipe/wrapper responsibility.
- Delayed-readout path matching still requires an unambiguous single post-scaler and does not choose the first branch in ambiguous topologies.

@coderabbitai please review
@gemini-code-assist please review

AllenYolk:
Addressed the latest actionable low-risk findings in `a4195340549d1988e65c2ce85f83adbf07586d1b`.

Implemented fixes:
- Changed the LTB FX hook insertion call from `call_module(module_name=...)` to positional `call_module(target, ...)`, matching `torch.fx.Graph.call_module` across PyTorch versions.
- Ensured `download_url()` closes the progress bar even if `requests.get()` fails before a response object is created, both for the non-resumable fallback and the ranged path.
- Added a regression test for progress-bar cleanup when the non-resumable fallback request fails.

Validation run:
- `uv run python -m pytest test/activation_based/test_ann2snn.py::TestDownloadUrl test/activation_based/test_ann2snn.py::TestLocalThresholdBalancingRecipe::test_recipe_converts_relu_to_channel_scalers_and_ifnode test/activation_based/test_ann2snn.py::TestLocalThresholdBalancingRecipe::test_recipe_uses_pre_spike_maxpooling -q` -> 9 passed
- `uv run python -m pytest test/activation_based/test_ann2snn.py test/activation_based/test_ann2snn_transformer.py -q` -> 195 passed
- `PYTHONPATH=$PWD uv run python -m sphinx -b html docs/source /tmp/spikingjelly-docs-pr-loop-check -W --keep-going` -> passed
- `git diff --check` -> passed

Intentionally not changed:
- I did not switch per-module restoration back to `module.train(training)`. Direct `module.training` assignment is still intentional for this cleanup path: it restores captured mixed train/eval flags in O(N), while recursively calling `train()` for every module is O(N^2) and repeatedly mutates descendants.

@coderabbitai please review
@gemini-code-assist please review

PR #709, opened 2026-07-22
Reduce Triton IF/LIF memory use when voltage sequences are not stored

### LINKED ISSUE #706: Avoid materialising unused Triton IF/LIF voltage sequences when store_v_seq=False
## Environment

- SpikingJelly baseline: `2797c5575515b59d7f09a3d5b732d6d25e148d13` (`master`)
- OS: Windows 11 10.0.26200
- Python: 3.11.15
- PyTorch: 2.7.1+cu128
- Triton: 3.3.1
- PyTorch CUDA runtime: 12.8
- Installed CUDA toolkit: 12.1 (`CUDA_PATH`)
- NVIDIA driver: 596.36
- GPU: NVIDIA GeForce RTX 4070 GPU, 8 GB, compute capability 8.9

## Problem

`IFNode` and `LIFNode` default to `store_v_seq=False`, which means callers only retain the final membrane potential. The standard Triton multi-step path nevertheless allocates a full `v_seq` tensor with the same `[T, N, ...]` shape as the input, writes every timestep into it, and then uses its final row to update the node state. Training also presents a full voltage-gradient sequence to the backward kernel even though only the final state is externally visible in this mode. This makes the default behaviour inconsistent with the intent of `store_v_seq=False` and adds allocation and memory traffic proportional to the full sequence length.

## Reproduction

The benchmark script is available on the proposed
[`perf/triton-last-state`](https://github.com/Magnushst/spikingjelly/tree/perf/triton-last-state)
branch. It uses a fixed seed of `20260717`, CUDA events, 5 warm-ups, 25 repetitions,
10 iterations per repetition, and `torch.cuda.max_memory_allocated()`:

```bash
python benchmark/benchmark_triton_last_state.py --output optimised.json
```

For a clean baseline/working-tree comparison:

```bash
git worktree add ../spikingjelly-baseline 2797c5575515b59d7f09a3d5b732d6d25e148d13
SJ_BENCH_SOURCE_ROOT=../spikingjelly-baseline \
  python benchmark/benchmark_triton_last_state.py --output baseline.json
python benchmark/benchmark_triton_last_state.py --output optimised.json
```

## Baseline evidence

Peak allocated memory for the default `store_v_seq=False` path:

| Mode | Shape / dtype | Baseline | Proposed | Reduction |
|---|---:|---:|---:|---:|
| Inference | `T=16, N=4096`, FP32 | 0.531 MiB | 0.281 MiB | 47.1% |
| Inference | `T=32, N=16381`, FP16 | 2.062 MiB | 1.062 MiB | 48.5% |
| Inference | `T=65, N=16381`, FP32 | 8.249 MiB | 4.187 MiB | 49.2% |
| Training | `T=16, N=4096`, FP32 | 1.532 MiB | 1.298 MiB | 15.3% |
| Training | `T=32, N=16381`, FP16 | 6.063 MiB | 5.095 MiB | 16.0% |
| Training | `T=65, N=16381`, FP32 | 24.498 MiB | 20.499 MiB | 16.3% |

IF and LIF produced the same peak-allocation figures. Four counter-ordered latency trials were noisy and mixed on this laptop, so this should be presented as a memory optimisation rather than as an end-to-end latency improvement.

For `T=65, N=16381`, FP32 IF training, a 10-iteration PyTorch Profiler capture showed:

- Forward custom-op device allocation: 127,779,840 to 85,841,920 bytes across 10 iterations.
- Forward kernel device time: 219.452 to 199.262 microseconds across 10 calls.
- Backward kernel device time: 219.868 to 132.890 microseconds across 10 calls.
- `aten::copy_` count: 20 to 10; the 10 device-to-device copies associated with retaining the final row disappeared.

## Proposed change

- Thread the existing node-level `store_v_seq` flag into standard IF/LIF Triton operators.
- Allocate a final-state-sized voltage output when the flag is false.
- Skip per-timestep voltage stores and write only the final state once.
- Seed the backward recurrence from the final-state gradient and skip per-timestep voltage-gradient loads in this mode.
- Preserve the existing full-sequence path when `store_v_seq=True`.
- Leave PLIF and mixed-precision neuron paths unchanged because their current contracts and intermediate requirements differ.

## Compatibility

- `IFNode`/`LIFNode` public behaviour and defaults are unchanged.
- The low-level `multistep_if`/`multistep_lif` helpers gain an optional `store_v_seq=True` argument, preserving existing callers and return behaviour by default.
- Torch, CuPy, PLIF, CPU, and mixed-precision paths are unchanged.
- No dependency or minimum-version change is required.
- The boolean is a Triton compile-time specialisation and an autotune-key component, so standard IF/LIF kernels can have separate full-sequence and final-state variants.

## Validation

- Compared full-sequence and final-state paths for IF and LIF.
- Covered inference and training, the default static-loop path (`T=1` and `T=7`), and the dynamic-loop path (`T=65`).
- Covered FP32, FP16, and BF16, hard and soft reset, and detach-reset.
- Compared spikes, final membrane state, and input gradients with dtype-appropriate tolerances.
- Exercised `torch.compile` with Inductor and checked fake-tensor output shapes.
- Benchmarked power-of-two and non-power-of-two feature counts with raw samples and peak allocation.

The complete relevant compile-compatibility module finished with 78 tests passed and 3 expected skips requiring a second CUDA device.

## Scope question

Would maintainers accept a focused PR that makes the standard Triton IF/LIF path honour `store_v_seq=False` at allocation and kernel level, while keeping full-sequence, mixed-precision, PLIF, and non-Triton behaviour unchanged?

Magnushst:
## Summary

This change makes the standard Triton IF/LIF multi-step path honour `store_v_seq=False` at allocation and kernel level. Instead of materialising a `[T, N, ...]` membrane-potential sequence and discarding all but its last row, the kernels write a final-state-sized output and the backward recurrence consumes only its final-state gradient.

## Motivation

`store_v_seq=False` is the node default and communicates that only the last membrane potential is retained. The previous Triton implementation still allocated and wrote the full voltage sequence. The redundant sequence accounted for almost half of measured inference peak allocation and roughly one sixth of training peak allocation in the tested cases.

## Root cause

The standard `multistep_if` and `multistep_lif` custom operators always allocated `v_seq` with `torch.empty_like(x_seq)`. Their Triton forward kernels stored `v` at every timestep, and the node selected/cloned `v_seq[-1]`. Autograd consequently exposed a full voltage-gradient output to backward even when the node only retained the last state.

## Changes

- Add `store_v_seq` as a compile-time flag and autotune key for standard IF/LIF forward and backward kernels.
- Allocate `v_init`-shaped output storage when only the final state is required.
- Skip per-timestep voltage stores and write the final state once.
- Initialise backward's voltage-gradient accumulator from the final-state gradient and skip sequence gradient loads in last-state mode.
- Update fake/meta registrations so spike, voltage, and intermediate shapes match eager execution.
- Thread the existing node flag through IF/LIF Triton dispatch.
- Preserve full-sequence behaviour by default in the low-level helpers and preserve mixed-precision/PLIF paths.
- Add a fixed-seed CUDA-event and peak-allocation benchmark.

## Correctness

The regression matrix compares `store_v_seq=True` with `False` for:

- IF and LIF;
- inference and training;
- static Triton loops (`T=1` or `T=7`) and the dynamic loop (`T=65`);
- FP32, FP16, and BF16;
- hard and soft reset;
- detach-reset in training;
- spike output, final membrane state, and input gradients;
- fake/meta output shapes and real Inductor forward/backward execution.

Tolerances are `atol=1e-5, rtol=1e-4` for FP32, `1e-3/1e-3` for FP16, and `1e-2/1e-2` for BF16. The two paths retain the same arithmetic recurrence; the wider low-precision tolerances match their storage precision.

Final compile-compat result: **78 passed, 3 skipped**. The skips require a second CUDA device.

## Performance

End-to-end latency was mixed and order-sensitive across four paired runs on the laptop GPU, so this PR does not claim a universal latency speedup. Median paired changes across the four trials ranged from a 28.5% slowdown to a 27.7% speedup depending on case. Raw samples, P25/P75 values, and negative results were retained and are available on request.

Kernel-level profiler evidence for IF training at `T=65, N=16381`, FP32 (10 iterations):

| Event | Baseline | Proposed | Change |
|---|---:|---:|---:|
| Forward Triton kernel device time | 219.452 us | 199.262 us | -9.2% |
| Backward Triton kernel device time | 219.868 us | 132.890 us | -39.6% |
| `aten::copy_` count | 20 | 10 | -10 calls |
| DtoD-copy count | 10 | 0 | -10 calls |

These profiler timings are one capture and are supporting evidence, not a cross-hardware speed claim.

## Memory

The memory result was identical across all four counter-ordered benchmark pairs and for IF/LIF:

| Mode | Shape / dtype | Baseline peak | Proposed peak | Reduction |
|---|---:|---:|---:|---:|
| Inference | `T=16, N=4096`, FP32 | 0.531 MiB | 0.281 MiB | 47.1% |
| Inference | `T=32, N=16381`, FP16 | 2.062 MiB | 1.062 MiB | 48.5% |
| Inference | `T=65, N=16381`, FP32 | 8.249 MiB | 4.187 MiB | 49.2% |
| Training | `T=16, N=4096`, FP32 | 1.532 MiB | 1.298 MiB | 15.3% |
| Training | `T=32, N=16381`, FP16 | 6.063 MiB | 5.095 MiB | 16.0% |
| Training | `T=65, N=16381`, FP32 | 24.498 MiB | 20.499 MiB | 16.3% |

## Benchmark methodology

- Baseline commit: `2797c5575515b59d7f09a3d5b732d6d25e148d13`
- GPU: NVIDIA GeForce RTX 4070 Laptop GPU, 8 GiB, driver 596.36
- Python 3.11.15; PyTorch 2.7.1+cu128; Triton 3.3.1
- Fixed seed `20260717`
- 5 warm-ups, 25 repetitions, 10 iterations per repetition
- CUDA events with synchronisation; compilation excluded after warm-up
- Median plus P25/P75 and all raw samples retained
- Four baseline/working-tree pairs in alternating order
- Peak delta from `torch.cuda.max_memory_allocated()` after allocator reset

Reproduce:

```bash
python benchmark/benchmark_triton_last_state.py --output optimised.json
```

Set `SJ_BENCH_SOURCE_ROOT` to a clean baseline worktree for the baseline run.

## Compatibility

- Node-level public API and default behaviour are unchanged.
- Low-level helpers keep full-sequence output by default through the optional `store_v_seq=True` parameter.
- Torch, CuPy, PLIF, CPU, and mixed-precision paths are unchanged.
- FP32, FP16, and BF16 were tested on CUDA compute capability 8.9.
- No dependency or minimum-version change.
- `store_v_seq` adds a compile-time/autotune specialisation dimension to the standard IF/LIF kernels.

## Testing

```text
uv format
ruff 0.15.22 format <six touched files>
ruff 0.15.22 check <six touched files>
PYTHONUTF8=1 python -m pytest test/activation_based/test_compile_compat.py -v
PYTHONUTF8=1 python -m pytest test/activation_based/test_triton_neuron.py -v
python benchmark/benchmark_triton_last_state.py --output <result.json>
sphinx-build -b html --keep-going docs/source <output-dir>
```

Results:

- Ruff: all checks passed; six task files formatted.
- Compile compatibility: 78 passed, 3 skipped. The skips require a second CUDA device.
- Shared Triton-neuron tests: 179 passed, 11 skipped in 452.70 seconds. Five skips require CuPy, and six pure-FP8-compute cases were rejected by the existing Triton capability probes on the test GPU.
- Normal Sphinx HTML build: succeeded.
- Strict `-W` Sphinx build generated the HTML but exited nonzero on two existing warnings in untouched `triton_kernel/torch2triton/torch2graph.py` docstrings.
- The Windows non-UTF-8 Inductor run initially failed with `UnicodeEncodeError`; the clean baseline reproduced the same failure. `PYTHONUTF8=1` made all three actual Inductor cases pass.

## Limitations

- Tested on one NVIDIA Ada GPU and Windows only.
- BF16 was tested; FP8/mixed-precision neuron operators were intentionally not changed.
- No Nsight Systems/Compute results were collected.
- End-to-end latency is noisy and not claimed as a guaranteed improvement.
- The complete repository-wide test suite was not run; the complete relevant compile-compatibility and shared Triton-neuron modules were run.

## Related issue

Closes #706.

## Summary by CodeRabbit

* **New Features**
  * Added a `store_v_seq` option for Triton-based multi-step IF/LIF to store only the final membrane potential (lower memory usage when disabled).
  * Introduced a Triton benchmark script to report latency and peak GPU memory across configurations.
* **Bug Fixes**
  * Improved consistency of spikes, gradients, and final neuron voltage across training and inference when toggling `store_v_seq`.
* **Tests**
  * Added Triton “last state” validation for training, inference, gradient equivalence, and fake-op output shapes.
* **Documentation**
  * Updated the changelog with the Triton IF/LIF memory optimization note.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,13 +78,13 @@ Module: `spikingjelly.activation_based.triton_kernel.neuron_kernel`.
 
 ### Improvements
 
-#### Triton Neuron Kernels
+#### Triton IF/LIF Memory Optimisation
 
 Module: `spikingjelly.activation_based.triton_kernel.neuron_kernel`.
 
 - Reduced memory usage for standard Triton IF and LIF neurons with
   `store_v_seq=False` by retaining only the final membrane potential instead of
-  materializing the full voltage sequence.
+  materialising the full voltage sequence.
 
 #### Distributed Training
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,14 @@ Module: `spikingjelly.activation_based.triton_kernel.neuron_kernel`.
 
 ### Improvements
 
+#### Triton Neuron Kernels
+
+Module: `spikingjelly.activation_based.triton_kernel.neuron_kernel`.
+
+- Reduced memory usage for standard Triton IF and LIF neurons with
+  `store_v_seq=False` by retaining only the final membrane potential instead of
+  materializing the full voltage sequence.
+
 #### Distributed Training
 
 Module: `spikingjelly.activation_based.distributed`.
--- a/benchmark/benchmark_triton_last_state.py
+++ b/benchmark/benchmark_triton_last_state.py
@@ -1,0 +1,160 @@
+import argparse
+import json
+import os
+import statistics
+import sys
+from pathlib import Path
+
+import torch
+import triton
+
+
+source_root = Path(
+    os.getenv("SJ_BENCH_SOURCE_ROOT", Path(__file__).resolve().parents[1])
+)
+sys.path.insert(0, str(source_root))
+
+from spikingjelly.activation_based.neuron.integrate_and_fire import IFNode
+from spikingjelly.activation_based.neuron.lif import LIFNode
+
+
+CASES = tuple(
+    (kind, mode, T, N, dtype)
+    for kind in ("if", "lif")
+    for mode in ("inference", "training")
+    for T, N, dtype in (
+        (16, 4096, torch.float32),
+        (32, 16381, torch.float16),
+        (65, 16381, torch.float32),
+    )
+)
+
+
+def percentile(values, q):
+    values = sorted(values)
+    index = (len(values) - 1) * q
+    lower = int(index)
+    upper = min(lower + 1, len(values) - 1)
+    weight = index - lower
+    return values[lower] * (1.0 - weight) + values[upper] * weight
+
+
+def make_node(kind):
+    cls = IFNode if kind == "if" else LIFNode
+    return cls(step_mode="m", backend="triton", store_v_seq=False).cuda()
+
+
+def run_case(
+    kind,
+    mode,
+    T,
+    N,
+    dtype,
+    warmup,
+    repetitions,
+    iterations_per_repetition,
+):
+    torch.manual_seed(20260717)
+    x = torch.randn(
+        T,
+        N,
+        device="cuda",
+        dtype=dtype,
+        requires_grad=mode == "training",
+    )
+    node = make_node(kind)
+
+    if mode == "training":
+
+        def step():
+            node.reset()
+            x.grad = None
+            spikes = node(x)
+            spikes.sum().backward()
+
+    else:
+
+        def step():
+            node.reset()
+            with torch.inference_mode():
+                node(x)
+
+    for _ in range(warmup):
+        step()
+    torch.cuda.synchronize()
+
+    start_event = torch.cuda.Event(enable_timing=True)
+    end_event = torch.cuda.Event(enable_timing=True)
+    samples_ms = []
+    for _ in range(repetitions):
+        start_event.record()
+        for _ in range(iterations_per_repetition):
+            step()
+        end_event.record()
+        end_event.synchronize()
+        samples_ms.append(
+            start_event.elapsed_time(end_event) / iterations_per_repetition
+        )
+
+    node.reset()
+    x.grad = None
+    torch.cuda.empty_cache()
+    torch.cuda.synchronize()
+    allocated_before = torch.cuda.memory_allocated()
+    torch.cuda.reset_peak_memory_stats()
+    step()
+    torch.cuda.synchronize()
+    peak_delta = torch.cuda.max_memory_allocated() - allocated_before
+
+    return {
+        "neuron": kind,
+        "mode": mode,
+        "T": T,
+        "N": N,
+        "dtype": str(dtype).removeprefix("torch."),
+        "warmup": warmup,
+        "repetitions": repetitions,
+        "iterations_per_repetition": iterations_per_repetition,
+        "median_ms": statistics.median(samples_ms),
+        "p25_ms": percentile(samples_ms, 0.25),
+        "p75_ms": percentile(samples_ms, 0.75),
+        "peak_delta_bytes": peak_delta,
+        "samples_ms": samples_ms,
+    }
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--warmup", type=int, default=5)
+    parser.add_argument("--repetitions", type=int, default=25)
+    parser.add_argument("--iterations-per-repetition", type=int, default=10)
+    parser.add_argument("--output", type=Path, required=True)
+    args = parser.parse_args()
+
+    results = []
+    for case in CASES:
+        result = run_case(
+            *case,
+            args.warmup,
+            args.repetitions,
+            args.iterations_per_repetition,
+        )
+        results.append(result)
+        print(
+            json.dumps(
+                {key: value for key, value in result.items() if key != "samples_ms"}
+            ),
+            flush=True,
+        )
+
+    payload = {
+        "torch": torch.__version__,
+        "triton": triton.__version__,
+        "gpu": torch.cuda.get_device_name(0),
+        "results": results,
+    }
+    args.output.write_text(json.dumps(payload, indent=2), encoding="utf-8")
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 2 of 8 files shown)


AllenYolk:
Thank you again for the thorough implementation and validation! Since these were small and focused follow-up changes, and maintainer edits were enabled, I pushed commit [`47da55eb`](https://github.com/fangwei123456/spikingjelly/commit/47da55eb30e05935a179f61a3adf9f1e3fc643b0) directly to the PR branch.

The commit:

- clones the stored final LIF state so that `MemoryModule.detach()` works with `store_v_seq=True`;
- adds a regression check for detaching both full-sequence and last-state nodes;
- adds the corresponding Change Log entry; and
- completes the bilingual `store_v_seq` documentation, including its default and output shape.

I reproduced the detach failure on the original PR revision on CUDA/Triton environment, then confirmed that the same test passes after the fix. The broader relevant test run completed with 235 passing tests.
LGTM. No further action is needed from you on these review points.

PR #710, opened 2026-07-26
refactor: reduce internal complexity across SNN modules

AllenYolk:
## Summary

- flatten ANN2SNN recipes and activation-based forwarding/neuron internals while keeping the converter/recipe architecture
- simplify distributed, precision, op-counter, CUDA, Triton, monitoring, profiling, learning, RNN, timing-based, and dataset implementations by removing redundant wrappers, state, branches, and compatibility layers
- document intentional public removals with direct migration paths; preserve mixed-precision kernels, dataset builders, legacy CuPy reference neurons, and leave FlexSN unchanged
- merge the current local `master` CUDA source-generation fix and retain its regression coverage

## Review triage

- accepted the OCR findings that removed public paths need explicit migration guidance, but did not restore forwarding wrappers or speculative compatibility branches
- retained Transformer Engine 2.16's current `te.autocast` API instead of adding a deprecated `fp8_autocast` fallback
- restored the accidental FlexSN exception-scope change because FlexSN was explicitly out of scope
- rejected the broad-diff concern because this branch is the accumulated, explicitly requested entropy-reduction refactor; the final diff removes substantially more code than it adds

## Validation

- local: `1521 passed, 347 skipped`
- g2 / A100 full regression before master merge: `1902 passed, 63 skipped`
- g2 / A100 post-merge CUDA/Triton regression: `252 passed, 28 skipped`
- g3 / RTX 4090 targeted regression: `357 passed, 7 skipped`
- Sphinx full rebuild: succeeded with no warnings
- changelog generation/check, formatting, and `git diff --check`: passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multi-step sequence forwarding with correct tuple outputs, recurrent final-state/dropout handling, and timing/shape edge cases; enhanced CUDA/CuPy and Triton kernel correctness and FP8-related stability.
* **New Features**
  * Added TD RMSNorm/SiLU + Transformer TD-equivalent recipes, Qwen2 offline calibration-driven conversion APIs, Transformer Engine FP8 precision mode, execution-plan based Triton neuron kernels, plus Gaussian population encoding, monitoring, online learning, and RNN support.
* **Breaking Changes**
  * Updated ANN→SNN conversion customization, distributed training APIs (enable flags → explicit mode), and precision/policy interfaces (stricter inputs; removed report); renamed Triton forward-plan API and dropped the Pydantic dependency.
* **Documentation**
  * Refreshed conversion, distributed-training, and API reference pages for the new interfaces.


AllenYolk:
## Code Review (Standards × Spec, two axes)

Method: reviewed the diff between merge-base `003ddc28` and branch `codex/entropy-reduction` (162 files, +5718/−12306) along two independent axes — **Standards** (repo `CONTRIBUTING.md`, a general agent code-editing guideline set R1–R5/§1–§12, and the Fowler smell baseline) and **Spec** (each claim in the PR body verified against the actual diff). All key findings below were spot-verified against the branch code before posting.

---

## Spec (claims verified against the diff)

**Verified claims** ✅

- FlexSN untouched (no FlexSN file appears in the diff)
- Kept TE 2.16's `te.autocast` API; no `fp8_autocast` fallback added (`precision/float8_te.py:556`)
- Mixed-precision kernels, dataset builders, and legacy CuPy reference kernels (`*_ptt`, `IFNodeFPTTKernel`, etc.) all preserved
- The master CUDA source-generation fix and its regression test are merged (`beccba27` + `229adc92`)
- Removed public symbols (`HookFactory`/`ReLURule`/`ThresholdOptimizer`/`TensorShardMemoryModule`/`SpikeSimEventEnergy*`/`build_eager_config`/`from_legacy`/dtensor facade) have no surviving references repo-wide; `CHANGELOG.md` and `docs/source/changelog.rst` are in sync

**🚫 Blocker**

1. **Violates "document intentional public removals with direct migration paths"**: the public attribute `BaseNode.lava_s_cale` was removed (`neuron/base_node.py:202` deletes `self.lava_s_cale = 1 << 6`), while `lava_exchange.py` replaced the 4 reads of `neu.lava_s_cale` with a hardcoded module constant `_LAVA_NEURON_SCALE = 1 << 6` (`lava_exchange.py:11, 815, 835, 1244, 1256`). `neuron.LIFNode().lava_s_cale` now raises `AttributeError`; models that customized this attribute will get **silently changed** Lava conversion results, and there is no CHANGELOG entry.
   → **Action**: restore the instance attribute and read the instance value in `lava_exchange`; or add a `lava_s_cale` removal entry with a migration path under "Breaking Changes and Notices".

**⚠️ Other findings**

2. **Undocumented scope creep**: the Summary lists only distributed/precision/op-counter/CUDA/Triton/monitoring/profiling/learning/RNN/dataset, but the diff also substantially rewrites `timing_based/encoding.py` (−452 lines), `timing_based/neuron.py` (−502 lines), `lava_exchange.py`, `quantize.py`, `encoding.py`, and `model/train_classify.py`. The changelog only mentions the `Tempotron` fix; the timing_based encoder rewrite (pydantic validation replaced with hand-written validation) is undocumented.
   → Update the Summary module list and add a changelog Notice for the timing_based rewrite.
3. **`pydantic` dependency silently removed** (`pyproject.toml:32`, `requirements.txt:10`, `docs/requirements.txt:7`), unmentioned in the Summary and changelog. Downstream projects relying on transitive pydantic installation will break.
   → Add a dependency-removal Notice to the changelog.
4. **Unnamed public class removals**: `ChannelVoltageHookFactory`, `ChannelWiseRateCodingReLURule`, `LocalThresholdBalancingHookFactory`, `LocalThresholdBalancingReLURule`, `ActivationRule` (ann2snn recipes), and `EagerParallelPolicy` (distributed) were all deleted with no surviving references, but the changelog only names the generic base classes. These are importable by module path even though not in `__all__`.
   → Add the names to the corresponding changelog entries (or note "including recipe-specific rule/factory classes").
5. Internal public helpers deleted/renamed without documentation (low risk): `ensure_cleanup_tmp_python_files`, `prepare_triton_neuron_forward_plan`, `resolve_sg_cupy_id_and_key`. No surviving references; optional, or fold into the changelog.
6. The claimed test counts (1521/1902/252/357 passed) cannot be verified from the diff; local CPU smoke tests passed (`test_encoding` 13 passed; `test_rnn`/`test_monitor`/`test_profiler`/`test_online_learning` 20 passed, 4 skipped).

---

## Standards (coding guidelines + smell baseline)

**What looks good** ✅

- Public contract changes handled per R2: every removed public API has a migration path in CHANGELOG "Breaking Changes and Notices", with docs updated in sync
- Behavior fixes and pure refactors listed separately in the changelog (Bug Fixes vs Breaking Changes), per R4 (e.g. stacked-RNN dropped final-layer state in `rnn.py`, tuple output in `functional/forward.py`, `Tempotron` squeeze losing a dim)
- Genuine duplication removal: `monitor.py`'s `_monitor_modules`/`_record` extraction (also fixing `remove_hooks` not calling `clear()`), `encoding.py`'s `_population_parameters`, `channel.py`'s `_ChannelShardConv` merge, `cuda_utils.py` object registries reduced from 5 mirrored dicts to 2 (§5 single source of truth)
- AST check of all changed files shows no dead imports

**Findings (by severity)**

1. **[Judgement, R4 staged migration / CONTRIBUTING "Keep PRs focused"]** A single diff mixes large-scale mechanical refactoring with semantic fixes (e.g. the behavior fixes in `rnn.py` and `timing_based/neuron.py` are interleaved with surrounding rewrites), making review hard to focus.
   → Per R4, split the behavior fixes into a preceding PR, or list a "files with behavior changes" manifest in the PR description.
2. **[Duplicated Code]** The local `_collect_tensors` in `op_counter/neuron_state.py:108` is an exact duplicate of the one this PR hoisted to `op_counter/base.py:44` (verified).
   → `from .base import _collect_tensors` and delete the local copy.
3. **[Middle Man]** `op_counter/neuromc/utils.py:5,14-15` merely re-exports `_infer_stage`/`_tensor_bits` from `..base` (verified in `__all__`); the only consumer, `neuromc/base_counter.py`, goes through this indirection.
   → Have `base_counter.py` import `from ..base import ...` directly and drop the two re-exports.
4. **[Judgement, §3/R3 fragile implicit invariant]** `ann2snn/recipes/step_mode_adapters.py`'s `_make_step_module` swaps `replacement.__class__ = step_type` on a deepcopy product, relying on the implicit convention that "layer wrapper classes add no state in `__init__` besides `step_mode`"; a future wrapper that adds initialized state will silently produce half-initialized objects.
   → Add a comment + post-construction assertion next to `_STEP_MODULE_TYPES` (e.g. verify `vars(replacement)` only gains `step_mode`), or go back to an explicit kwargs-construction table.
5. **[Judgement, Duplicated Code]** The new `_td_module_from_ann` in `ann2snn/operators.py` overlaps with `_make_step_module` in Linear/Conv2d parameter-copy logic.
   → Share a parameter-copy helper; acceptable to keep if the two paths are genuinely semantically different.
6. **[Minor, §2/§12]** `neuron/lif_variants.py:141-142`: `param_shape` and `random_shape` are identical duplicate variables (verified).
   → Delete `random_shape`, use `param_shape` uniformly.
7. **[Judgement, §3]** `timing_based/neuron.py:83-85` uses a bare `nn.Module()` as a namespace to preserve the checkpoint key `model.summation_layer.weight` — the comment explains why and the contract is preserved, but readability suffers; consider a named private submodule class. Non-blocking.

**Unverified items / residual risks (R5)**

- No GPU/full dependencies in this environment; numeric-path changes (IF/LIF/PLIF cupy kernel parameter sinking, `rnn.py` state advancement, `Tempotron` rewrite, encoding's `linspace` replacing the hand-rolled loop) need before/after comparison via CI/GPU regression.
- `encoding.py` incidentally fixed a division-by-zero in the old code when `pop_dim=1` (linspace semantics); not in the changelog — suggest adding an entry.
- `base_node.py` removed the `hasattr(torch, "compile")` guard and TypeError fallback; `channel.py` removed the `dist.is_initialized()` guard. The direction matches §4 (fail early on internal invariants) but is a behavior tightening — suggest confirming in the PR description that the minimum supported torch version supports the `options=` parameter.

---

**Summary**: Standards axis — 7 findings (worst: one diff mixing mechanical refactoring with semantic fixes, against R4 staged migration); Spec axis — 1 Blocker (public attribute `lava_s_cale` removed undocumented, silently changing Lava conversion for models that customized it) + 5 findings (worst: pydantic dependency silently removed). Recommend addressing at least the Blocker and the dependency-removal changelog entry before merging.

CHANGELOG.md
```diff
@@ -177,8 +208,27 @@ Module: `spikingjelly.activation_based.distributed`.
   throughput, peak-memory, and tensor-parallel debug fields; legacy
   FSDP2-specific communication counter keys are no longer emitted.
 
+#### Timing-Based Models
```
AllenYolk: should be placed near "timing-based neurons"

CHANGELOG.md
AllenYolk, line 1: Changelog should be simplified. The current version is too verbose.

spikingjelly/activation_based/ann2snn/examples/cnn_mnist.py
```diff
@@ -315,22 +310,6 @@ def main(args):
         drop_last=False,
     )
 
-    # loss_function = nn.CrossEntropyLoss()
```
AllenYolk, line 318: Should not be removed. It specifies how the cnn is yielded and can be used to train the CNN

spikingjelly/activation_based/ann2snn/__init__.py
AllenYolk, line 1: Besides the tests, all ann2snn examples should be checked in CUDA environment

spikingjelly/activation_based/distributed/__init__.py
AllenYolk, line 1: Distributed benchmarks should be checked in cuda environment besides the tests.

spikingjelly/activation_based/examples/memopt/train_distributed.py
AllenYolk, line 1: should run the script in cuda env besides those tests

spikingjelly/activation_based/layer/attention.py
```diff
@@ -365,9 +341,8 @@ def backend(self):
     @backend.setter
     def backend(self, value: str):
         self._backend = value
-        self.qkv_lif.backend = value
-        self.attn_lif.backend = value
-        self.proj_lif.backend = value
+        for module in (self.qkv_lif, self.attn_lif, self.proj_lif):
+            module.backend = value
```
AllenYolk: Seem to be useless. functional.set_backend will recursively set the backend of all submodules.

spikingjelly/activation_based/layer/attention.py
```diff
@@ -529,9 +504,8 @@ def backend(self):
     @backend.setter
     def backend(self, value: str):
         self._backend = value
-        self.qk_lif.backend = value
-        self.attn_lif.backend = value
-        self.proj_lif.backend = value
+        for module in (self.qk_lif, self.attn_lif, self.proj_lif):
+            module.backend = value
```
AllenYolk: The same as above.

spikingjelly/activation_based/learning.py
```diff
@@ -9,6 +9,20 @@
...
+        monitor.InputMonitor(synapse, function_on_input=torch.Tensor.detach),
+        monitor.OutputMonitor(sn, function_on_output=torch.Tensor.detach),
+    )
+
+
+def _write_weight_gradient(synapse, delta_w):
+    if synapse.weight.grad is None:
+        synapse.weight.grad = -delta_w
+    else:
+        synapse.weight.grad = synapse.weight.grad - delta_w
```
AllenYolk: "Write" is weird. Use "accumulate" or some other word.

spikingjelly/activation_based/learning.py
```diff
@@ -9,6 +9,20 @@
 from . import neuron, monitor, base
 
 
+def _create_spike_monitors(synapse, sn):
+    return (
+        monitor.InputMonitor(synapse, function_on_input=torch.Tensor.detach),
+        monitor.OutputMonitor(sn, function_on_output=torch.Tensor.detach),
+    )
```
AllenYolk: Unnecessary wrapper. Make it inline.

spikingjelly/activation_based/monitor.py
```diff
@@ -50,13 +50,23 @@ def __getitem__(self, i):
...
             raise ValueError(i)
 
+    def _monitor_modules(self, net, instance):
+        if instance is None:
+            instance = type(net)
+        for name, module in net.named_modules():
+            if isinstance(module, instance):
+                self.monitored_layers.append(name)
+                self.name_records_index[name] = []
+                yield name, module
```
AllenYolk: Weird. A generator should not modify self's state. Consider move row 62, 63 outside the generator, or remove the wrapper.

spikingjelly/activation_based/neuron/integrate_and_fire.py
```diff
@@ -402,8 +368,6 @@ def multi_step_forward(self, x_seq: torch.Tensor):
                     v_reset=self.v_reset,
                     detach_reset=self.detach_reset,
                     surrogate_function=self.surrogate_function,
-                    forward_kernel=self.forward_kernel,
-                    backward_kernel=self.backward_kernel,
                 )
```
AllenYolk, line 371: forward_kernel and backward_kernel cache the cupy kernels. Are these kernels cached in the current version? How are they cached?

spikingjelly/activation_based/precision/capability.py
```diff
@@ -40,12 +40,7 @@ def _transformer_engine_fp8_status() -> tuple[
             {},
         )
 
-    if hasattr(te, "autocast"):
-        autocast_api = "autocast"
-    elif hasattr(te, "fp8_autocast"):
-        autocast_api = "fp8_autocast"
-    else:
-        autocast_api = None
+    autocast_api = "autocast" if hasattr(te, "autocast") else None
```
AllenYolk, line 43: What about fp8_autocast? Is it an alias of autocast?

spikingjelly/datasets/utils.py
```diff
@@ -581,56 +555,11 @@ def integrate_events_segment_to_frame(
...
-    # mask.append(np.logical_not(mask[0]))
-    # for i in range(2):
-    #     position = events['y'][mask[i]] * width + events['x'][mask[i]]
-    #     events_number_per_pos = np.bincount(position)
-    #     idx = np.arange(events_number_per_pos.size)
-    #     frames[i][idx] += events_number_per_pos
-    # frames = frames.reshape(2, height, width)
-    # print('correct accumulation by bincount\n', frames)
-
-    frame = np.zeros(shape=[2, H * W])
```
AllenYolk, line 622: Don't remove these. Summarize these instead.

CHANGELOG.md
AllenYolk: Addressed in f8e49068. The Timing-Based Models improvement now directly follows the Timing-Based Neurons bug-fix subsection.

CHANGELOG.md
AllenYolk: Addressed in f8e49068. I condensed the Unreleased changelog substantially, keeping user-visible capabilities and compatibility notices while removing implementation-level narration.

spikingjelly/activation_based/ann2snn/examples/cnn_mnist.py
AllenYolk: Addressed in f8e49068. I restored a self-contained commented training recipe, including model construction, optimizer, ten-epoch training, checkpoint saving, and validation.

spikingjelly/activation_based/ann2snn/__init__.py
AllenYolk: Addressed and validated in f8e49068. I ran every runnable ANN-to-SNN example on CUDA (MNIST, CIFAR-10 ResNet, ImageNet ResNet, ViT/STA, BERT TD-equivalent, and RoBERTa/SpikeZIP). That validation also found and fixed two real example issues: FX iteration over Proxy shapes and an unconvertible functional average-pool call.

spikingjelly/activation_based/distributed/__init__.py
AllenYolk: Addressed and validated in f8e49068. CUDA benchmark modes none, DP, TP, FSDP2, FSDP2+TP, and PP were exercised with NCCL_P2P_DISABLE=1 and NCCL_IB_DISABLE=1. PP exposed a real bug: automatic stage costs were partitioned independently on differently loaded ranks. Rank 0 is now the single layout authority; automatic PP then passed for both CIFAR10-DVS VGG and Spikformer.

spikingjelly/activation_based/examples/memopt/train_distributed.py
AllenYolk: Addressed by validation. I ran the actual memopt train_distributed.py workflow on g2 CUDA for one complete CIFAR10-DVS epoch (T=16, batch=32, torch backend, memopt level 1): train 29.86%, validation 39.30%, 58.60 samples/s.

spikingjelly/activation_based/layer/attention.py
AllenYolk: Addressed in f8e49068. I removed the parent backend proxy state and setter; recursive functional.set_backend now updates the neuron children directly.

spikingjelly/activation_based/layer/attention.py
AllenYolk: Addressed in f8e49068 together with the equivalent SpikingSelfAttention code.

spikingjelly/activation_based/learning.py
AllenYolk: Addressed in f8e49068. Renamed the helper to _accumulate_weight_gradient.

spikingjelly/activation_based/learning.py
AllenYolk: Addressed in f8e49068. The monitor construction is now inline in each learner; the one-line wrapper is gone.

spikingjelly/activation_based/monitor.py
AllenYolk: Addressed in f8e49068. _monitor_modules now eagerly builds and returns a list, then updates monitor state immediately; iteration no longer triggers hidden mutations.

spikingjelly/activation_based/neuron/integrate_and_fire.py
AllenYolk: Confirmed: the CuPy kernels remain cached. Single-step IF lazily stores and reuses forward_kernel/backward_kernel on each neuron instance. Multi-step IF uses module-level _IF_FWD_KERNEL_CACHE and _IF_BWD_KERNEL_CACHE keyed by kernel attributes in cuda_kernel/auto_cuda/neuron_kernel/integrate_and_fire.py. No code change was needed.

spikingjelly/activation_based/precision/capability.py
AllenYolk: Checked against current NVIDIA Transformer Engine documentation. fp8_autocast is the deprecated compatibility entry point, not an alias contract we should rely on; autocast(enabled=..., recipe=...) is the current API. The implementation intentionally keeps te.autocast without an extra fallback branch.

spikingjelly/datasets/utils.py
AllenYolk: Addressed in f8e49068. I kept the essential reason as a compact comment: indexed += can lose repeated positions, while bincount accumulates every event.

PR #711, opened 2026-07-26
feat(functional): add explicit state transition APIs

AllenYolk:
## Summary

- Add 36 flat public tensor-state update functions under `spikingjelly.activation_based.functional` for neurons, delay and synaptic filters, and STDP-family learning traces.
- Migrate official modules with standard, non-hookable state equations to call these functions while preserving `module(x)`, `reset()`, `detach()`, memory attributes, `step_mode`, `backend`, and output behavior.
- Keep hookable/customizable Node execution paths unchanged in this phase; their standard equations are exposed without forcing subclasses through them.
- Use the existing Inductor equations as the single implementation of IF/LIF/PLIF multi-step state transitions, backed by a bounded, PID-aware process-local compilation cache.
- Keep backend selection in `MemoryModule`; backend-specific functions are named explicitly with `_cupy`, `_triton`, or `_inductor`.
- Add bounded-cache and FlexSN registry diagnostics needed to inspect the new execution paths.

## API scope and semantics

- Only modules with clear memory and state-transition semantics receive functional APIs. TD forwarding, STA scheduling helpers, Dropout masks, NeuNorm, and SpikeZIP bias scheduling remain module implementation details.
- `*_step` performs one complete time-step transition and returns outputs together with the next state.
- `*_multi_step` is published only where the sequence path has an independent, nontrivial implementation. A plain Python loop over `*_step` remains in the module instead of becoming another public wrapper.
- STDP functions group traces as `(trace_pre, trace_post)` and return `(update, trace_next)`; convolution padding and module configuration remain owned by the module adapter.
- SpikeZIP exposes the STBIF state equation, but not its known-horizon bias allocation as a countdown state machine.
- All public functions are directly reachable from the `functional` namespace.

## Compatibility

The public `MemoryModule` contract is unchanged. Existing calls, state management, backend selection, step modes, outputs, and user-defined `MemoryModule` subclasses retain their behavior. The BNTT exceptional path also restores the previous state correctly.

## Validation

- Broad affected-suite baseline: 437 passed, 85 skipped.
- Final focused suites: learning 13 passed; SpikeZIP 38 passed, 7 skipped; combined focused checks 16 passed.
- CPU `torch.compile` smoke tests and bounded-cache stress tests passed.
- Targeted Ruff checks, `git diff --check`, and changelog consistency checks passed.
- `make clean && make html` and strict Sphinx (`-W --keep-going`) both completed with zero warnings.

Real CUDA/CuPy/Triton kernel execution was not available on the macOS host; dispatch and fake-kernel coverage was exercised instead.

## Summary by CodeRabbit

* **New Features**
  * Added functional, explicit state-transition APIs for neuron, layer, and learning operations.
  * Added backend-specific single-step and multi-step neuron support across available acceleration backends.
  * Added functional STDP, mSTDP, and reward-modulated learning updates.
  * Added diagnostic visibility for FlexSN kernel registry references.

* **Improvements**
  * Improved compiled neuron execution by sharing bounded, process-aware caches across compatible modules.

* **Documentation**
  * Expanded API references with state-handling conventions, backend naming, and functional learning guidance.


AllenYolk:
## Code Review (Standards × Spec, two axes)

Method: reviewed the diff between merge-base `003ddc28` and branch `codex/native-functional-state` (42 files, +11143/−745) along two independent axes — **Standards** (repo `CONTRIBUTING.md`, a general agent code-editing guideline set R1–R5/§1–§12, and the Fowler smell baseline) and **Spec** (each claim in the PR body verified against the actual diff). All key findings below were spot-verified against the branch code before posting.

---

## Spec (claims verified against the diff)

**Verified claims** ✅

- Migration is behavior-preserving: spot-checked Delay (`layer/misc.py:371` uses `self.queue[:] = queue_next` to preserve queue identity), MaskedPSN (preserves the "update queue, then raise OverflowError" exception side effect, `functional/neuron.py:3882-3935`), SlidingPSN, SynapseFilter, GatedLIF, STBIF, StatefulEncoder/LatencyEncoder/WeightedPhaseEncoder, learning wrappers, and SpikeZIP/STA recipes — all faithful
- Hookable Nodes genuinely not rewired (as declared): IFNode's training path still goes through `super().multi_step_forward`, cupy/triton still call kernels directly; `nonlinear_if.py` (QIF/EIF) and `adapt.py` (Izhikevich) completely untouched
- Named backend interfaces exist (`*_cupy`/`*_triton`/`*_inductor`) with backend selection kept in modules
- Inductor cache is bounded (`_MAX_ENTRIES = 64`, LRU eviction + fork reset)
- TD ANN wrappers excluded (`ann2snn/operators.py` not in the diff)
- Documentation in place: 5 new `docs/source/APIs/spikingjelly.activation_based.functional*.rst` files, CHANGELOG +44 lines

**⚠️ Findings (actionable)**

1. **Function count mismatch (79 ≠ 80)**: the five new files' `__all__` total neuron 52 + layer 6 + encoding 3 + learning 7 + ann2snn 11 = **79** (verified per-file via AST).
   → Correct "80" in the PR body, or add the missing API.
2. **Single/multi-step coverage policy contradicts the implementation, literally**: `functional/neuron.py:1635` `if_multi_step` (also `lif_multi_step`:2188, `plif_multi_step`:2653) is exactly a Python loop over the single-step function, while the spec says "only single-step is exposed when the module multi-step path is exactly the inherited loop". Defensible (IF/LIF/PLIF also have independent cupy/triton/inductor multi-step semantics), but the wording and implementation disagree.
   → Document the exception in the PR body/docstrings, or drop the pure-loop `*_multi_step` functions.
3. **Undisclosed exception-side-effect change (BNTT)**: `functional/layer.py:255` `batch_norm_through_time_single_step` restores `track_running_stats` via try/finally; the old code did **not** restore it when `f(x)` raised. An improvement, but it contradicts the overall "preserve exception-side-effect behavior" promise and is undisclosed.
   → Note it in the changelog.
4. **NeuNorm memory type change**: `layer/bn.py:314-319` adds an explicit tensor-materialization branch for float `self.x`; behavior is equivalent but the path differs.
   → Add a float-init regression test assertion.
5. **Scope creep**: the diff contains `beccba27` "fix(cuda): remove RST from kernel strings" and FlexSN registry diagnostics (`triton_kernel/flexsn/custom_ops.py` +42, etc.). Recorded in the CHANGELOG but not mentioned in the PR Summary (the PR claims to be based on master tip `beccba27`; relative to the merge-base these changes fall inside this PR's diff).
   → Add a note to the Summary or split them out.
6. **STBIF buffer in-place semantics change**: the old implementation used `self.cur_output.zero_()` + boolean-index in-place buffer updates; the new one replaces the tensor wholesale (`neuron/spikezip.py:179-186`), invalidating external code holding the old buffer reference.
   → Confirm no external references exist, or note it in the changelog.

**Unverified items**: no torch in this environment — `419 passed, 13 skipped`, the Inductor compile smoke, and the cache stress test could not be rerun; the CuPy/Triton GPU paths are self-declared as unverified.

---

## Standards (coding guidelines + smell baseline)

**What looks good** ✅

- `encoding.py`, `layer/misc.py`, `layer/container.py`, `layer/bn.py`, `learning.py`, `neuron/psn.py`, `neuron/spikezip.py` all delegate to the functional functions — single source of truth achieved there
- Parameter order/return conventions are consistent across functions; multi-step uniformly returns `(spike_seq, v_next, v_seq_or_none)`
- R2 satisfied: new public APIs ship with rst docs and CHANGELOG entries; tests are module-vs-function equivalence tests, per §9

**Findings (by severity)**

1. **[Top concern, §5/R3 single source of truth]** Core neuron classes (hookable Nodes such as IF/LIF) do not delegate to the new functional functions, so the same LIF transition exists in multiple parallel implementations: `neuron/lif.py:365 _eval_multi_step_forward`, `functional/neuron.py:2188 lif_multi_step`, `neuron/inductor_cache.py _build_lif_multi_step_graph` (plus cupy/triton kernels). **Note: the PR explicitly declares this as a phased decision ("hookable Node implementations are not rewired in this phase"), which the Spec axis accepts as compliant**; however, ~50 functional neuron functions currently have zero production callers (only tests call them, §2 concept count), and the parallel implementations risk drifting in the interim.
   → Provide a tracking mechanism for the follow-up phase (issue/follow-up PR) and add long-term anti-drift notes to the module-vs-function equivalence tests for IF/LIF; or state explicitly whether the end state is modules delegating to functions or the reverse.
2. **[Fragile implementation, verified]** `neuron/inductor_cache.py:62` `surrogate_key` uses `id()` as the cache key for non-builtin surrogates: `return ("surrogate_instance", id(surrogate_function))`. While an entry is alive, the closure reference prevents id-reuse collisions, but that invariant no longer holds after bounded-LRU eviction, and every new instance creates a new key, causing cache churn.
   → Return `None` (no caching) for non-spikingjelly surrogates, or use a weakref-based key, and document the invariant.
3. **[Fragile implementation, verified]** `neuron/base_node.py:383` matches exceptions by string: `if str(exc) == "backend='inductor' requires torch.compile."` (§3 pointless re-wrapping + coupling to message text).
   → Pass the node name into `inductor_cache.compile_graph`, raise with the name at that site, and delete the try/except in base_node.
4. **[§3 pointless wrapping]** `functional/neuron.py`'s `if_multi_step_triton` (L3624) and the `*_cupy` family (L2861–3483) merely forward to kernels + reshape, with no production callers.
   → Delete them, or have the modules' backend branches call them (linked to the end-state decision in finding 1).
5. **[§5/§7 global state]** `inductor_cache` introduces a process-level global `_COMPILED_GRAPHS` singleton and runs `os.register_at_fork` at import time (§7: avoid import side effects); `_MAX_ENTRIES = 64` is a magic number; `clear()`/`info()` mainly serve tests (§2: interfaces exposed only for tests). Also: when `surrogate_key is None`, `cache_key=None` and every forward re-runs `torch.compile`.
   → At minimum, document ownership and invalidation conditions, and mark `clear()` as test-only.
6. **[§8 boilerplate docstrings]** Every function carries a ~50-line docstring repeating the same "this function does not read module state…" paragraph. Bilingual docs match CONTRIBUTING, but the boilerplate belongs in the module docstring once, with function-level docstrings keeping only the deltas (about 2/3 of the 4441 lines are docstrings).
7. **[§1/§8]** `tools/generate_memorymodule_inventory.py` (397 lines) contains a hardcoded `CLASSIFICATION_BY_NAME` with one-off migration-plan codenames like `"2c_function_only"` and `"migrate_after_cache_prototype"`, producing a one-off inventory markdown.
   → Move it to the PR description or a wiki instead of committing it.
8. **[Minor]** `layer/misc.py` `Delay.single_step_forward` does an O(delay) `tuple(self.queue)` copy + slice assignment per step with no measured justification (§1); `functional/encoding.py:433` uses `getattr(self, "alpha", None)` with a default masking a missing attribute — prefer plain `self.alpha` (§4 fail early).

---

**Summary**: Spec axis — no Blocker, 6 actionable findings (best first fixes: the 79≠80 API count and the single/multi-step policy wording — two claim-vs-implementation mismatches); Standards axis — 8 findings (worst: hookable Nodes not delegating, leaving multiple parallel implementations of the same transition — a declared phased decision, but the end state needs tracking to prevent drift; plus two verified fragile implementations: the `id()` cache key and the string-matched exception). Overall direction and documentation quality are good; recommend addressing the findings before merging.

spikingjelly/activation_based/functional/ann2snn.py
```diff
@@ -0,0 +1,343 @@
...
+    "spikezip_matmul_delta",
+    "spikezip_matmul_sequence_delta",
+    "sta_spike_encoder_single_step",
+]
+
+
+def spikezip_release_bias_single_step(
+    y: torch.Tensor,
+    bias: torch.Tensor | None,
+    realize_time: int,
```
AllenYolk: `realize_time` here seems to be unnecessary.

spikingjelly/activation_based/functional/ann2snn.py
```diff
@@ -0,0 +1,343 @@
...
+    if bias is None or realize_time <= 0:
+        return y, realize_time, False
+    bias = bias.to(device=y.device, dtype=y.dtype).view(bias_view_shape)
+    return y + bias / bias_steps, realize_time - 1, True
+
+
+def spikezip_release_bias_multi_step(
+    y_seq: torch.Tensor,
+    bias: torch.Tensor | None,
+    realize_time: int,
```
AllenYolk: `realize_time` here seems to be unnecessary

spikingjelly/activation_based/functional/ann2snn.py
```diff
@@ -0,0 +1,343 @@
...
+    released_steps = min(y_seq.shape[0], realize_time)
+    if bias is None or released_steps <= 0:
+        return y_seq, realize_time, 0
+    bias = bias.to(device=y_seq.device, dtype=y_seq.dtype).view(bias_view_shape)
+    y_next = y_seq.clone()
+    y_next[:released_steps] = y_next[:released_steps] + bias / bias_steps
+    return y_next, realize_time - released_steps, released_steps
+
+
+def spikezip_matmul_delta(
```
AllenYolk: `spikezip_matmul_delta` and `spikezip_matmul_seq_delta` do not seem like state transition functions. 

They should be either modeled as state transition functions here (a.k.a., `[*inputs, *states] -> [*outputs, *updated_states]`), or modeled as vanilla mappings but placed elsewhere.

spikingjelly/activation_based/functional/ann2snn.py
```diff
@@ -0,0 +1,343 @@
...
+    if transpose_b:
+        b_cum = b_cum.transpose(-2, -1)
+    y_cum = a_cum @ b_cum
+    y_seq = torch.empty_like(y_cum)
+    y_seq[0] = y_cum[0]
+    y_seq[1:] = y_cum[1:] - y_cum[:-1]
+    return y_seq
+
+
+def sta_spike_encoder_single_step(
```
AllenYolk: Why not place this function in sta_transformer.py?

spikingjelly/activation_based/functional/neuron.py
AllenYolk, line 1: Charge and reset functions are unnecessary. They are so short that wrapping them seems to be redundant.

spikingjelly/activation_based/functional/layer.py
```diff
@@ -0,0 +1,190 @@
...
+    if not isinstance(delay_steps, int) or delay_steps < 0:
+        raise ValueError("delay_steps must be a non-negative integer")
+
+    queue_with_x = (*queue, x)
+    if len(queue_with_x) > delay_steps:
+        return queue_with_x[0], queue_with_x[1:]
+    return torch.zeros_like(x), queue_with_x
+
+
+def neunorm_single_step(
```
AllenYolk: No need to wrap NeuNorm as a function.

spikingjelly/activation_based/neuron/inductor_cache.py
AllenYolk, line 1: Too complex, and seem to have overlap in functionality with neuron state transition functions.

PR #721, opened 2026-08-09
Optimize PLIF memory use and FlexSN compiler integration

AllenYolk:
## Summary

- add a reproducible single-GPU SNN benchmark runner and compile-boundary probe for SpikingVGG, SEW-ResNet, Spikformer, LIF, and FlexSN
- remove PLIF backward's redundant `[v_init; v_seq]` materialization by passing the initial voltage and voltage sequence separately to the Triton kernels
- align the five FlexSN operators with the existing `torch.library.triton_op` + `wrap_triton` path while preserving the opaque `custom_op` fallback
- preserve public operator schemas, state/reset behavior, output layout, and registered-autograd contracts

## Measured evidence

Experiments used an NVIDIA A100-SXM4-80GB, PyTorch 2.7.1+cu118, CUDA 11.8, and Triton 3.3.1.

### PLIF

- `aten::cat` calls: 8 -> 0
- peak allocated memory: approximately 10.42% lower across the three tested shapes
- median kernel latency: 5.31% to 9.40% lower across the three tested shapes
- one small-shape round regressed slightly, so this is presented as a memory-materialization optimization rather than a significant latency or end-to-end speedup

### FlexSN

The corrected training probe uses `requires_grad=(True, True)` and rejects unavailable Triton training kernels instead of silently measuring the eager fallback.

| Path | Inference median | Training median |
| --- | ---: | ---: |
| Torch compile | 0.096256 ms | 0.495616 ms |
| registered Triton FlexSN | 0.098304 ms | 0.473088 ms |
| opaque FlexSN | 0.193536 ms | 1.032192 ms |

Registered Triton is approximately 1.97x faster than opaque inference and 2.18x faster than opaque training in this probe. Generated training code directly launches the FlexSN forward/backward Triton kernels and contains no runtime `torch.ops.sj` FlexSN call. All compared paths use one Dynamo graph with zero graph breaks and recompiles.

Kernel activity count is retained only as structural evidence, not as the optimization target. Inference has the same four physical activities for registered and opaque FlexSN despite the latency difference.

## Validation

- `python -m pytest -q test/activation_based`: 1448 passed, 341 skipped
- `python -m pytest -q benchmark/test/test_snn_single_gpu_benchmark.py`: 7 passed
- g2 CUDA FlexSN/benchmark suite: 135 passed
- corrected A100 compile probe: six inference/training rows passed with matching outputs and gradients
- `ruff format --check`, `ruff check`, `py_compile`, and `git diff --check` passed

## Scope

This change does not add model-name routing, a new runtime backend, a feature flag, or a `torch.autograd.Function`. It does not modify multi-GPU, data/optimizer, AMP policy, CUDA Graph, sparse compute, or Inductor/cuDNN lowering. PyTorch version evidence currently covers 2.7.1; the supported version matrix should be rechecked in CI.

## Summary by CodeRabbit

- **New Features**
  - Added configurable single-GPU benchmarking for inference and training, with latency, memory, profiling, GPU monitoring, and JSON reports.
  - Added benchmark comparison workflows across source versions, including paired results and failure handling.
  - Added compile-boundary probing across supported model cases with structured success, unsupported, and diagnostic results.

- **Bug Fixes**
  - Improved PLIF training and mixed-precision backward handling, including initial-state gradient propagation.
  - Improved FlexSN operator registration and compatibility.

- **Tests**
  - Expanded CUDA coverage for FlexSN and PLIF correctness and compatibility.

- **Documentation**
  - Updated FlexSN documentation to reflect registered operator integration.

### DIFF
--- a/benchmark/benchmark_snn_single_gpu.py
+++ b/benchmark/benchmark_snn_single_gpu.py
@@ -505,6 +505,7 @@ def run_case(args: argparse.Namespace) -> dict[str, Any]:
     ).resolve()
     sys.path.insert(0, str(source_root))
     import spikingjelly
+    from spikingjelly.activation_based import functional
 
     package_file = Path(spikingjelly.__file__).resolve()
     try:
@@ -541,8 +542,6 @@ def run_case(args: argparse.Namespace) -> dict[str, Any]:
     criterion = torch.nn.CrossEntropyLoss()
 
     def reset_net() -> None:
-        from spikingjelly.activation_based import functional
-
         functional.reset_net(state_model)
 
     def step() -> torch.Tensor:
--- a/benchmark/benchmark_snn_single_gpu.py
+++ b/benchmark/benchmark_snn_single_gpu.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import argparse
+from contextlib import suppress
 import json
 import os
 import platform
@@ -23,24 +24,18 @@ PHASES = ("inference", "training")
 EXECUTIONS = ("eager", "compile")
 
 
-def percentile(values: list[float], q: float) -> float:
-    if not values:
-        raise ValueError("percentile requires at least one value")
-    ordered = sorted(values)
-    index = (len(ordered) - 1) * q
-    lower = int(index)
-    upper = min(lower + 1, len(ordered) - 1)
-    weight = index - lower
-    return ordered[lower] * (1.0 - weight) + ordered[upper] * weight
-
-
 def summarize_samples(samples_ms: list[float]) -> dict[str, float]:
     if not samples_ms:
         raise ValueError("samples_ms must not be empty")
+    deciles = (
+        statistics.quantiles(samples_ms, n=10, method="inclusive")
+        if len(samples_ms) > 1
+        else samples_ms * 9
+    )
     return {
         "median_ms": statistics.median(samples_ms),
-        "p10_ms": percentile(samples_ms, 0.10),
-        "p90_ms": percentile(samples_ms, 0.90),
+        "p10_ms": deciles[0],
+        "p90_ms": deciles[8],
         "stdev_ms": statistics.stdev(samples_ms) if len(samples_ms) > 1 else 0.0,
     }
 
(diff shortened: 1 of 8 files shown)


AllenYolk:
Addressed the actionable review findings in `ebabb543`:

- clamped both PLIF backward history offsets and added CUDA `T=1` coverage;
- filtered credential-like environment variables from benchmark metadata;
- made monitor cleanup terminate/kill/close safely;
- added per-case matrix/probe timeouts, partial failure records, diagnostics, and process-group cleanup;
- paired benchmark comparisons only across rounds where both sources succeeded;
- made FlexSN probe availability checks phase-specific and recorded registration mode for every result;
- removed the redundant `opcheck` assertion.

Intentionally not applied:

- `channels_last_3d`: `(T,N,C,H,W)` uses dimension 1 as batch, not channel; `channels_last_3d` would encode the wrong semantic layout. The current flatten-to-per-frame-NCHW conversion is intentional.
- inherited `PYTHONPATH`: replacing it with the selected source root is required for baseline/candidate import isolation.
- hidden-child cache fallback: `--child` is an internal parent-owned protocol and generated-code inspection requires its isolated cache.
- relaxed bool assertions: those tests intentionally enforce JSON boolean schema, not general truthiness.
- `detach()` on saved PLIF values: it still shares the version counter and does not solve the claimed aliasing issue.

Validation after the fixes:

- local `test/activation_based`: `1448 passed, 341 skipped`;
- local benchmark/FlexSN targeted suite: `134 passed, 8 skipped`;
- A100 PLIF review suite: `18 passed` including `T=1/4/32`, static and dynamic kernels;
- A100 compile probe: `6/6` rows `ok`, one graph, zero breaks/recompiles, matching outputs and gradients, expected registered/opaque codegen;
- Ruff, `py_compile`, and `git diff --check` passed.

A final A100 rerun was attempted after the last tooling-only hardening, but all authorized A100s were occupied, so it exited immediately rather than queueing or interfering with active jobs.

AllenYolk:
@gemini-code-assist review

AllenYolk:
@coderabbitai review

AllenYolk:
Addressed the new review batch in `203155f8`.

Applied:
- tolerate `nvidia-smi dmon` exit races and close its log on every failed startup path;
- reject zero-latency measurements instead of dividing by zero, and remove redundant round sorting;
- count only generated Triton/C++/extern kernel call sites, excluding unrelated `.run()` helpers;
- rename the two PLIF tests to match their actual `aten::cat` and mp-helper coverage.

Not applied:
- the two duplicate schema findings: both `triton_op` and `custom_op` modes were executed locally and render the same `SymInt` schema;
- wrapping LIF construction: kernel failures occur during compile/forward and are already returned with their original exception;
- a Torch-version assertion for Dynamo counters: current reads are guarded, while pinning a version would undermine this cross-version probe.

Validation: benchmark tests `17 passed`; FlexSN file `120 passed, 8 skipped`; Triton-neuron file `22 passed, 119 skipped`; schema test passed in both registration modes; Ruff, format check, `py_compile`, and `git diff --check` passed. No review threads were resolved and no review bot was manually retriggered.

PR #744, opened 2026-09-03
Fix MSTDPLearner initialising eligibility without a dtype

tritsystem:
## What

`MSTDPLearner.step()` initialises `self.eligibility` with `torch.zeros(...)` and
an explicit `device=` but **no `dtype=`**, so it is always `float32`. On a
half-precision synapse (AMP), the resulting weight gradient is `float32` and
`weight.grad = -delta_w` raises:

```
RuntimeError: attempting to assign a gradient with dtype 'float' to a tensor
with grad_dtype 'Half'.
```

`float32` (default) and `float64` work — `float64` only because
`reward (float64) * eligibility (float32)` promotes up to `float64`.
`half * float32` promotes to `float32`, so half breaks.

`MSTDPETLearner` already does this correctly with
`torch.zeros_like(self.synapse.weight)`; this change brings `MSTDPLearner` in
line.

## Reproduce (on current `master`)

```python
import torch, torch.nn as nn
from spikingjelly.activation_based import learning, neuron

fc = nn.Linear(4, 3, bias=False).half()
sn = neuron.LIFNode(step_mode="s")
lr = learning.MSTDPLearner(step_mode="s", synapse=fc, sn=sn,
                           batch_size=2, tau_pre=2., tau_post=2.)
sn(fc((torch.rand(2, 4) > 0.5).half()))
lr.step(torch.ones(2).half())          # RuntimeError before this change
```

## Fix

```diff
                 self.eligibility = torch.zeros(
                     self.batch_size,
                     *self.synapse.weight.shape,
                     device=self.synapse.weight.device,
+                    dtype=self.synapse.weight.dtype,
                 )
```

`torch.zeros_like` isn't usable here because of the leading `batch_size` dim, so
`dtype=` is added as an explicit kwarg matching the existing `device=`.

## Tests

Adds `test_mstdp_learner_eligibility_follows_synapse_dtype` to
`test/activation_based/test_learning.py`, parametrised over `float16` / `float64`.
`float16` fails on `master`, passes with this change; `float64` passes both ways.
Full `test/activation_based/test_learning.py` is green (18 passed);
`uvx ruff@0.15.22 check` is clean.

---

*Developed with AI assistance; I reproduced the bug, verified the fix and the
regression against `master`, and reviewed every line.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mixed-precision learning support by ensuring eligibility tensors and reward-based updates match the synapse’s data type and device.
  * Preserved expected gradient behavior for `float16` and `float64` synapses.
* **Documentation**
  * Updated learning documentation and changelogs to clarify device and data type handling.


AllenYolk:
Maintainer follow-up: aligned tensor rewards with the synapse dtype/device, updated the existing mixed-dtype test and changelog, and verified all required checks pass. Thanks!

PR #745, opened 2026-09-03
Add regression test: neuron state and output follow module dtype

tritsystem:
## What

Parametrised regression coverage for the state/dtype invariant behind the recent
`MemoryModule._apply` fix (#743) and `MSTDPLearner` eligibility fix (#744): a
neuron's recurrent state and its floating forward output must follow the module
dtype after `.to(...)`.

Over 14 single-step `BaseNode` subclasses (`IFNode`, `LIFNode`,
`ParametricLIFNode`, `QIFNode`, `EIFNode`, `IzhikevichNode`, `KLIFNode`,
`ComplementaryLIFNode`, `ILIFNode`, `OTTTLIFNode`, `SLTTLIFNode`,
`HalfThresholdIFNode`, `ActivationAwareIFNode`, `SimpleIFNode`):

- after `.to(float64)` / `.to(float16)`: a floating forward output has the module
  dtype, and every floating registered memory has the module dtype after
  `reset()`;
- `reset()` forgets the batch size (no stale batch dim on the next forward at a
  different batch).

All 42 cases pass on `master` -- this is pure regression coverage, no behaviour
change.

`DSR*` (multi-step only) and `CUBALIFNode` (distinct state contract) are already
exercised in `test_functional_neuron.py` / `test_neuron_variants.py` and are out
of scope here.

## Tests

`test/activation_based/test_neuron_dtype_invariance.py` -- 42 passed;
`uvx ruff@0.15.22 check` + `format --check` clean; runs green alongside the
existing neuron test files (95 passed together).

---

*Developed with AI assistance; I ran and reviewed every line.*

## Summary by CodeRabbit

* **Tests**
  * Added coverage ensuring single-step neuron outputs and floating-point state match the module’s configured dtype.
  * Added coverage confirming neuron state resets correctly when the batch size changes between forward passes.

### DIFF
--- a/test/activation_based/test_neuron_dtype_invariance.py
+++ b/test/activation_based/test_neuron_dtype_invariance.py
@@ -1,0 +1,59 @@
+import pytest
+import torch
+
+from spikingjelly.activation_based import neuron
+
+# Concrete BaseNode subclasses that construct with no required arguments.
+# Single-step neurons that take a plain ``(batch, features)`` input.
+# (DSR* are multi-step only; CUBALIFNode has a distinct state contract --
+# both are exercised in test_functional_neuron.py / test_neuron_variants.py.)
+# The intent is coverage: a neuron's recurrent state and its
+# (floating) forward output must follow the module's dtype after ``.to(...)`` --
+# the same invariant class as #743 (MemoryModule._apply) and #744 (MSTDPLearner).
+NO_ARG_NODES = [
+    neuron.IFNode,
+    neuron.LIFNode,
+    neuron.ParametricLIFNode,
+    neuron.QIFNode,
+    neuron.EIFNode,
+    neuron.IzhikevichNode,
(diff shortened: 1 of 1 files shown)


AllenYolk:
Thanks for the contribution. This broad matrix duplicates invariants already covered by `MemoryModule`/`BaseNode`, while its state-dtype assertion does not exercise tensor memories. We prefer focused regression tests for concrete neuron-specific bugs, so we are closing this without merging.

tritsystem:
Understood, thank you for the clarification — that's a fair call. A broad general-invariant matrix isn't the right shape of test to contribute here; I'll aim for focused regression tests tied to a concrete, confirmed neuron-specific bug going forward, and make sure any state-dtype coverage actually exercises tensor memories rather than just the scalar/named-memory case this one missed. Appreciate you taking the time to explain the reasoning rather than just closing it.

PR #754, opened 2026-09-15
fix(ottt): detach neuron state between timesteps in ottt_online_training

### LINKED ISSUE #593: Online Training Through Time (OTTT) Training Function Error
**Issue type**

- [x] Bug Report
- [ ] Feature Request
- [ ] Help wanted
- [ ] Other

**SpikingJelly version**

`0.0.0.0.15`

**Description**

When trying to train a network with OTTT, the function runs into an error, seemingly because the neuron state is not detached from the computational graph after each time-step. Adding `functional.detach_net(model)` inside `functional.ottt_online_training` eliminated the error for me.

**Code to reproduce the error**

```python
import torch
from torch import nn
from torch.nn import functional as F

from spikingjelly.activation_based import neuron, layer, functional

net = layer.OTTTSequential(
    nn.Linear(8, 4),
    neuron.OTTTLIFNode(),
    nn.Linear(4, 2),
    neuron.LIFNode()
)

optimizer = torch.optim.SGD(net.parameters(), lr=0.1)

T = 4
N = 2
online = True
for epoch in range(2):

    x_seq = torch.rand([N, T, 8])
    target_seq = torch.rand([N, T, 2])

    functional.ottt_online_training(model=net, optimizer=optimizer, x_seq=x_seq, target_seq=target_seq, f_loss_t=F.mse_loss, online=online)
    functional.reset_net(net)
```

**Error**
```
RuntimeError: Trying to backward through the graph a second time (or directly access saved tensors after they have already been freed). Saved intermediate values of the graph are freed when you call .backward() or autograd.grad(). Specify retain_graph=True if you need to backward through the graph a second time or if you need to access saved tensors after calling backward.
```

Likith-Yadav:
## Summary

`ottt_online_training` runs `loss.backward()` at every time step, but it never detaches neuron state. Membrane potentials / traces stay attached to the previous graph, so the next step raises:

```
RuntimeError: Trying to backward through the graph a second time
```

That is the crash in #593 (and in the docstring example). `fptt_online_training` already calls `detach_net` after each step; this PR does the same after the OTTT backward (and after the online `optimizer.step()`).

OTTT/SLTT are not BPTT: each time step should only use the current-step graph. Detaching is required for both `online=True` and `online=False`.

If the last layer is `OTTTLIFNode`, training forward returns `[spike, trace]`. The loss is taken on the spike tensor so that setup trains as well (the example in the docs uses `LIFNode` as the last layer; both are covered by tests).

Fixes #593.

## Testing

```
pytest test/activation_based/test_online_learning.py
```

14 passed, including a regression test for the #593 snippet (`LIFNode` last layer) and the all-`OTTTLIFNode` variant, with `online=True` and `online=False`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved online training across multiple time steps by detaching model state between updates and preventing repeated backpropagation errors.
  * Fixed handling of spike-and-trace outputs during loss calculation.
  * Corrected sequential processing for consecutive OTTTLIFNode layers to preserve state and avoid duplicate updates.

* **Tests**
  * Added coverage for multi-timestep training in online and non-online modes.
  * Verified compatibility with OTTTLIFNode and LIFNode outputs, including stacked OTTTLIFNode layers.

* **Documentation**
  * Documented the updated online training behavior.


Likith-Yadav:
@felixonmars , @chanokin , @piernov could you check out my pr and approve the workflow?

Likith-Yadav:
Thanks for catching this. `OTTTSequential`'s parameterless path does apply the next `OTTTLIFNode` to both spike and trace, so the output is nested `[[spike, trace], [spike, trace]]` and `y_t[0]` was still a list.

Pushed a follow-up that walks the first element until a tensor remains (covers both a single `[spike, trace]` pair and the nested case). Added a regression test with two consecutive `OTTTLIFNode` layers.

AllenYolk:
Thanks for the PR. Sorry for the late response!

The `detach_net(model)` change is correct and should be kept. However, I don’t think CodeRabbit’s recursive output-unwrapping suggestion is the right fix. The nested output is caused by `OTTTSequential`. When the next module is an `OTTTLIFNode`, `OTTTSequential` should call it once with the spike tensor and use its returned `[spike, trace]` as the new pair, rather than applying the same stateful node separately to both spike and trace. Would you be willing to update `OTTTSequential` and remove the recursive unwrapping?

AllenYolk:
We are also preparing **a broader refactor of SpikingJelly’s online-learning modules**. The current OTTT/SLTT implementation follows the original two-pass design and runs each weight layer once with spikes and again with traces; see [OTTT's source code](https://github.com/pkuxmq/OTTT-SNN). The refactor will use spikes in the forward pass and traces directly in the backward pass, eliminating the redundant computation. It will also improve usability and provide a unified framework for multiple online-learning methods. Please stay tuned!!

Likith-Yadav:
Thanks for the review, and sorry for following the unwrap suggestion too closely.

I kept `detach_net(model)`. `OTTTSequential` now calls `OTTTLIFNode` once on the spike tensor and uses the returned `[spike, trace]` pair as the new state, instead of applying the same stateful node to both spike and trace. The recursive unwrap helper is gone; `ottt_online_training` still takes `y_t[0]` when the last layer returns a one-level `[spike, trace]` pair.

Happy to adjust further if anything else looks off.

AllenYolk:
Thanks for the update! This follows the intended approach and looks good to me. I pushed a small maintainer cleanup: inlined the two helpers, replaced the class-name check with `isinstance` to support subclasses, and removed the unnecessary fallback. Thanks for the contribution!

PR #755, opened 2026-09-17
fix(precision): make _resolve_device_type return the actual device type

li-lizhe:
## Summary

`_resolve_device_type()` in `capability.py` uses string-matching to
classify devices: if the string starts with "cuda" → "cuda", with "mps"
→ "mps", otherwise "cpu".  This mislabels every other accelerator
(NPU, XPU, `privateuse1` backends) as "cpu", which:

1.  Reports the wrong `device_type` in the capability report.
2.  Falls into the CPU bf16 detection path instead of probing the
    actual accelerator.

## Change

- **`_resolve_device_type`** — use `torch.device(device).type` which
  returns the canonical type string for any device (`"cuda"`, `"mps"`,
  `"cpu"`, `"npu"`, `"xpu"`, `"privateuse1"`, ...).
- **bf16 detection path** — added an explicit `device_type == "cpu"`
  branch, with a new `_probe_device_bf16()` helper for unknown
  accelerators.
- **`_validate_fp16` / `_validate_bf16`** — restructured to handle
  non-CUDA, non-MPS accelerators with honest error messages instead of
  falling through to CUDA-only checks.

No `.cuda()` or `if torch_npu` branches are added — this is purely
device-agnostic.

## Verification

Changed file lints clean.  The only test that uses the report
(`test_capability_report_splits_can_convert_and_can_execute`) passes
with `device="cpu"` (unchanged behaviour).  CUDA, MPS and CPU paths
are untouched — only the default `else` fallthrough changes.

## Related

Continues the same device-agnostic direction as #748 (remove `.cuda()`
hardcoding in DSpike search network).

## Summary by CodeRabbit

* **Bug Fixes**
  * Capability reports now preserve the requested device type, including unsupported accelerators, instead of incorrectly identifying them as CPU.
  * Improved bfloat16 capability reporting and validation for CUDA and other device types.
  * FP16 validation now clearly rejects unsupported device types with an error.
* **Documentation**
  * Updated the changelogs to document the precision capability reporting fixes.


AllenYolk:
Thanks for the PR. Using `torch.device(device).type` is the correct fix.

I suggest removing `_probe_device_bf16()`. Tensor allocation does not reliably represent AMP support, and a generic capability probe adds unnecessary complexity. For now, unknown device types should be reported correctly but treated as unsupported for fp16/bf16 until the backend is explicitly implemented and tested.

